PR #7, opened 2025-07-17
Add `packaging` to main dependencies

bbatsell:
Recent changes to `cache.py` introduced a dependency on the `packaging` package. It is in the `uv.lock` because it is depended on by `pytest`, but if the user does not install the dev dependencies then it does not get installed. The README-recommended `uvx` run command thus results in an import error.


daaain:
Oh! Thanks a lot 😸

PR #38, opened 2025-11-04
Fix Unicode escape in tool use content rendering

tim-watcha:
## Problem

When exporting Claude Code transcripts to HTML, non-ASCII characters (Korean, Chinese, Japanese, etc.) in tool use content were being displayed as escaped Unicode sequences like `\uba3c\uc800` instead of their actual characters.

Example of the issue:
```json
{
  "thought": "\uba3c\uc800 /airflow/schema/rds/ \ub514\ub809\ud1a0\ub9ac \uad6c\uc870\ub97c \uc815\ub9ac\ud574\ubcf4\uaca0\uc2b5\ub2c8\ub2e4."
}
```

Should display as: "먼저 /airflow/schema/rds/ 디렉토리 구조를 정리해보겠습니다."

## Root Cause

The `json.dumps()` function in `renderer.py` was using the default `ensure_ascii=True` parameter, which escapes all non-ASCII characters.

## Solution

Added `ensure_ascii=False` parameter to `json.dumps()` calls:
- `renderer.py:296` - Tool use content rendering (critical fix)
- `generate_style_guide.py:375` - Style guide generation (consistency)

## Testing

All existing tests pass without modification:
- 229 unit tests passed
- No breaking changes
- HTML charset is properly set to UTF-8
- File encoding is properly configured

## Additional Verification

Verified that:
- HTML templates include `<meta charset='UTF-8'>`
- All file writes use `encoding="utf-8"`
- `html.escape()` preserves Unicode characters correctly
- Jinja2 autoescape only escapes HTML tags, not Unicode

## Regenerating Existing HTML Files

To regenerate previously exported HTML files with the fix applied, users need to run:
```bash
uvx claude-code-log --clear-cache
uvx claude-code-log --clear-html
```

This will clear the cache and remove existing HTML files, allowing them to be regenerated with proper Unicode rendering.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of non-ASCII characters in rendered content and generated outputs so they display correctly rather than as escaped Unicode sequences.

### DIFF
--- a/claude_code_log/renderer.py
+++ b/claude_code_log/renderer.py
@@ -327,7 +327,7 @@ def render_params_table(params: Dict[str, Any]) -> str:
         # If value is structured (dict/list), render as JSON
         if isinstance(value, (dict, list)):
             try:
-                formatted_value = json.dumps(value, indent=2)  # type: ignore[arg-type]
+                formatted_value = json.dumps(value, indent=2, ensure_ascii=False)  # type: ignore[arg-type]
                 escaped_value = escape_html(formatted_value)
 
                 # Make long structured values collapsible
--- a/scripts/generate_style_guide.py
+++ b/scripts/generate_style_guide.py
@@ -372,7 +372,7 @@ def generate_style_guide():
         # Write style guide data
         with open(jsonl_file, "w") as f:
             for entry in style_guide_data:
-                f.write(json.dumps(entry) + "\n")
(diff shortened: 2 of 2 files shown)


tim-watcha:
@daaain Done!

daaain:
Released in https://github.com/daaain/claude-code-log/releases/tag/0.8.0

PR #48, opened 2025-11-25
Fix Pygments Lexer Performance Bottleneck

cboos:
(maybe it's just noticeable on Windows with moderately large logs, but still...)

## Summary

Fix critical performance bottleneck in syntax highlighting that caused 2+ second delays per message on Windows systems. Achieves **40-125x speedup** by eliminating filesystem I/O from lexer selection.

## Problem

Message rendering was experiencing severe slowdowns when processing Read tool results with file paths. Investigation revealed:

1. **Root cause**: Pygments' `get_lexer_for_filename()` performs filesystem I/O operations (stat, file existence checks, byte reading) to determine the appropriate lexer
2. **Windows impact**: With antivirus scanning, this causes 100-1000ms delays per file path, even when the file no longer exists or has changed since the transcript was recorded
3. **Secondary issues**:
   - Double Pygments calls for Read tool results (once for full content, once for preview)
   - Slow lexer pattern matching: iterating through all 909 filename patterns using fnmatch for every file

### 1. Filename Pattern Cache (main fix)

Use Pygments' public `get_all_lexers()` API to build a reverse filename pattern → lexer alias lookup table:

- Built once on first use, cached as function attribute
- Supports all 597 Pygments lexers automatically (909 filename patterns)
- Eliminates ALL filesystem I/O by pattern matching against basenames only
- Correct business logic: render time shouldn't depend on current filesystem state

### 2. Extension-Based Fast Path

Added O(1) extension lookup for common patterns:

- Simple `*.ext` patterns indexed by extension (e.g., "py" → "python" lexer)
- Falls back to full pattern matching only when extension lookup fails
- Reduces lexer lookup from 50ms to ~10ms

### 3. Eliminate Double Highlighting

For Read tool results with >12 lines, we were calling Pygments twice (full content + preview). Now:

- Extract preview HTML from already-highlighted content using regex
- Find first 5 `<tr>` rows and close unclosed tags
- Eliminates redundant highlighting pass

### 4. Timing Instrumentation Module

Added `renderer_timings.py` module for performance profiling:

- Controlled by `CLAUDE_CODE_LOG_DEBUG_TIMING` environment variable
- Context managers for phase timing and operation statistics
- Reports slowest messages and operations
- Useful for future performance investigations

## Performance Results

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| Slow messages | 2136ms | 53ms | **40x** |
| Lexer lookup | 50ms | 10ms | **5x** |
| Read tool result | 120ms | 44ms | **2.7x** |

Fast case (messages without file paths): 0.9ms → 0.9ms (unchanged)

## Test Plan

- [x] All 253 unit tests passing
- [x] Type checking passes (pyright: 0 errors)
- [x] Manual testing with large transcripts on Windows
- [x] Verified timing output with `CLAUDE_CODE_LOG_DEBUG_TIMING=1`
- [x] Confirmed lexer selection matches original behavior for common file types

## Files Changed

- `claude_code_log/renderer.py` - Main performance fixes
- `claude_code_log/renderer_timings.py` - New timing utilities module

## Summary by CodeRabbit

* **Performance Optimization**
  * Faster syntax highlighting via cached lexer selection and reduced I/O; overall rendering and initialization speed improvements.

* **Improvements**
  * Richer tool-result rendering with clearer headers, deduplication, and contextual linking.
  * Better session organization, navigation entries, token-usage summaries, and message pairing/hierarchy for clearer presentation.

* **New Features**
  * Collapsible truncated code previews for large edits with full-content expandable view.

* **Documentation**
  * Updated fold-state diagram and transition table for clearer UI state explanation.

* **Tests**
  * Added regression tests for preview truncation and collapsible previews.


cboos:
Hello Daniel,

> I have quite a chunky .claude directory (100s of MBs of JSONL logs) ... rewrite the cache to use SQLite instead of JSON files. 

I think my claude-code-log usage is more limited than yours, so I can't really tell the pros/cons here.
I typically use claude-code-log targeting one session only (one .jsonl file) and then read/explore the HTML transcript for that one, then eventually store it somewhere for documentation/reference. So, I've not yet seen or benefited from the cache (my shortcut actually does `uv run claude-code-log --clear-html --no-cache $path -o $html`!). Maybe I'm not using it the best possible way.

> What's been holding me back is a lack of good integration tests: another thing I really want to do is a full fake .claude directory as test fixture that has all weird edge cases so we can run the entire processing pipeline and maybe even do snapshots on the HTML output.

Interesting, yes. Multiple such `.claude/projects` would be good: also at least one with Windows style paths, and "versioning" them (`test/test_data/projects/v2.0.41-win/[.claude/projects/]...`).

> That said, this seems to be orthogontal because while it's also filesystem based, but it's more working around a specific Pygments lookup quirk rather than this library's own filesystem access, so sounds good!

Indeed, this one has nothing to do with the cache. I try to no introduce slowdowns with my changes, so I was really worried when I had this unexpected slowdown. Fortunately, a fix/workaround was possible.
 
> I was just wondering if we could somehow turn on these timings in CI and print results into [job summaries](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#adding-a-job-summary) in a similar way as [code coverage](https://github.com/daaain/claude-code-log/actions/runs/19685372643#summary-56389505364) so we can track how timings change over time?

Good idea, I'll see what I can do!

cboos:
> > I was just wondering if we could somehow turn on these timings in CI and print results into job summaries...
> Good idea, I'll see what I can do!

Well, for now and the .jsonl we have in the repo, it's not that interesting.
What we could do here is a test that first generates a large .jsonl in a deterministic way, process it, and we measure that.

PR #56, opened 2025-12-02
Fix cross-session tool pairing on session resume

cboos:
When session B resumes session A, both JSONL files may contain messages with the same tool_use_id. The pairing logic was using only tool_use_id as the dictionary key, causing tool_use from session A to be incorrectly paired with tool_result from session B.

Changed _identify_message_pairs() and _reorder_paired_messages() to use (session_id, tool_use_id) tuple keys instead of just tool_use_id, ensuring tool pairs are matched within the same session only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed session isolation to correctly track tool interactions within individual sessions when conversations are resumed.


daaain (approved):
Wow, that was an obscure one 😅

PR #57, opened 2025-12-02
Filter out warmup messages + parse IDE tags for concise display in summaries

daaain:
Summary of Implementation

  1. Warmup Session Filtering

  Sessions where all user messages are "Warmup" are now completely
  hidden:
  - Filtered from cache updates in converter.py
  - Skipped when generating individual session HTML files
  - Filtered from the main HTML transcript rendering

  2. IDE Compact Indicators in Session Previews

  Verbose IDE tags at the start of messages are replaced with compact
  emoji indicators:

  | IDE Tag           | Compact Indicator     |
  |-------------------|-----------------------|
  | <ide_opened_file> | 📎 /full/path/to/file |
  | <ide_selection>   | ✂️ /full/path/to/file |
  | <ide_diagnostics> | 🩺 diagnostics        |

  Key design decisions:
  - Only processes IDE tags at the start of content (not embedded in
  quoted JSONL)
  - Shows full file path (not just filename)
  - Handles files without extensions (e.g., justfile)
  - Applies compact indicators before truncation to 1000 chars

  Files Modified

  - claude_code_log/utils.py - Core logic for warmup detection and
  compact IDE indicators
  - claude_code_log/converter.py - Warmup filtering in cache and file
  generation
  - claude_code_log/renderer.py - Warmup filtering in HTML generation
  - test/test_utils.py - Comprehensive tests for all new functionality

## Summary by CodeRabbit

* **New Features**
  * Warmup-only sessions excluded from lists, caches, exports; agent-only (empty-user) sessions omitted from navigation and indexes.
  * Hook summaries rendered as collapsible summaries with commands, outputs, and error details; large command outputs get previews with expand/collapse.
  * Session previews compact IDE tags and better init-command descriptions; Bash inputs/outputs treated as “tool” messages.
  * Session navigation includes per-session fields and aggregated token-usage labeling.

* **Style**
  * Theming variables, hook UI styles, refined Bash styling, and local-timestamp rendering script added.

* **Tests**
  * New tests for warmup filtering, IDE compaction, previews, and hook summary rendering.

### DIFF
--- a/claude_code_log/renderer.py
+++ b/claude_code_log/renderer.py
@@ -2449,7 +2449,7 @@ def _process_bash_output(text_content: str) -> tuple[str, str, str, str]:
         re.DOTALL,
     )
 
-    output_parts: List[str] = []
+    output_parts: List[tuple[str, str, int, str]] = []
     total_lines = 0
 
     if stdout_match:
@@ -2475,7 +2475,7 @@ def _process_bash_output(text_content: str) -> tuple[str, str, str, str]:
     if output_parts:
         # Build the HTML parts
         html_parts: List[str] = []
-        for output_type, escaped_content, line_count, raw_content in output_parts:
+        for output_type, escaped_content, _, _ in output_parts:
             css_name = f"bash-{output_type}"
             html_parts.append(f"<pre class='{css_name}'>{escaped_content}</pre>")
 


daaain:
I also noticed loads of these:

```sh
Line 428 of /Users/dain/.claude/projects/-Users-dain-workspace-claude-code-log/b6a1d63c-a734-48ce-ad58-1acf7c7c2aca.jsonl | 1 validation error for SystemTranscriptEntry
content
  Field required [type=missing, input_value={'parentUuid': '3397d941-...45a7-af43-c5f037af42a7'}, input_type=dict]
```

Which turned out to be stop hooks, so added support for them.

And then also noticed that I had a fair few sessions where I did `/clear` and `! just test` (or similar Bash command) which rendered correctly in the session summary (dropping clear and showing the shell command), but in the actual session missed the whole bash bit:

image

So fixed these too 😅 these were actually really tricky because they were in the HTML, just filtered out. Here's Claude's summary:

I've fixed the bash messages not rendering issue. The problem was in
  the JavaScript filter logic in transcript.html:

  Root Cause: The "Tool" filter only expanded to tool_use and
  tool_result CSS classes, but bash commands rendered by the Bash tool
  have classes bash-input, bash-stdout, and bash-stderr. Since these
  weren't included in any filter type, messages with these classes were
  hidden by default when the filter was applied.

  Fix: Updated the filter expansion in transcript.html to include bash
  message types:
  1. Filter expansion (line 325-326): Added bash-input, bash-stdout,
  bash-stderr to the tool filter
  2. Message counts (line 301): Updated tool count selector to include
  bash messages
  3. Visible counts (line 398-399): Updated visible tool count selector
  to include bash messages

  Now bash commands will properly show/hide when toggling the "Tool"
  filter, consistent with other tool types.

So now we get:

image

daaain:
> As I use claude directly from the CLI less often these days, I didn't see that these were missing... good catch! And I'll right-align those (commands typed by the user, so belonging to the user's side of the conversation).

I have a love / hate relationship with the terminal version, I much more prefer the VS Code one's UI and lack of scroll jank, but it's lagging behind so much with features...

Anyway, thanks a lot for picking this up! I think once you're happy let's merge and we can cut a 0.9 version?

cboos:
> Anyway, thanks a lot for picking this up! I think once you're happy let's merge and we can cut a 0.9 version?

This one is good to go.

However, before a 0.9, I still have another "polish" PR cooking, with a few more important visual fixes. The last commit there is in the works (apparently, even if visually a "simple" glitch, this needed a lots of thinking on Claude's part; let's hope there will be no regressions...).

PR #106, opened 2026-04-01
Fix search broken when HTML saved with different filename

moshethebeadle:
## Summary
- The `isIndexPage` detection in search relied on `window.location.pathname` matching specific URL patterns (`index.html`, `projects/`, etc.)
- When the generated HTML is saved/downloaded with a different filename (e.g. browser "Save As"), the detection fails and search indexes zero items, completely breaking search functionality
- Replaced URL-based detection with DOM-based detection (`document.querySelectorAll('.project-card').length > 0`), which correctly identifies the page type regardless of filename or serving context

## Test plan
- [ ] Generate an index page and verify search works normally
- [ ] Save the index page with a different filename (e.g. "My Projects.html") and verify search still works
- [ ] Verify search still works on transcript pages (non-index pages)
- [ ] Run existing snapshot tests (snapshot will need updating)

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable page detection for search so index pages show grouped project/session results and transcript pages highlight messages in-place; result-panel display, navigation, and transcript linking behave correctly.
  * Transcript pages now re-index when filters change, ensuring search results stay up to date and only when applicable.

* **Style**
  * Minor snapshot/CSS adjustments affecting error/display formatting.

### DIFF
--- a/claude_code_log/html/templates/components/search.html
+++ b/claude_code_log/html/templates/components/search.html
@@ -88,6 +88,11 @@
 
         // Initialize UI state
         updateSearchInputState();
+
+        // Re-index when filters change (transcript pages only)
+        if (!searchState.isIndexPage) {
+            setupTranscriptFilterObserver();
+        }
     }
 
     // Build search index from current page
@@ -733,8 +738,7 @@
     }
 
     // Re-index when filters change (for transcript pages)
-    if (!searchState.isIndexPage) {
-        // Listen for filter changes
+    function setupTranscriptFilterObserver() {
         let isUpdatingFilters = false;
         const observer = new MutationObserver((mutations) => {
             // Only react to filtered-hidden class changes, not search highlights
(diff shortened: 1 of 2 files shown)


daaain:
Hey @moshethebeadle, thanks a lot for this fix!

I had a look and all the other usage of `isIndexPage` happens after (or invoked by) `initSearch` except the `if (!searchState.isIndexPage)` the bunny highlighted, so once you fix that we should be good.

I approved the tests and the snapshots will need to be updated once you're ready to get them green.

cboos:
I apologize for the confusion with Git. I didn't intend to push to the branch `moshethebeadle:fix/isIndexPage-detection`, so I created PR #108. However, it seems that the command `gh pr checkout` set things up in a way that led me here anyway.  And Claude Code is just too trigger happy with `push` these days ;-)

I tested everything locally, and it appears to be working fine.

PR #108, opened 2026-04-11
Fix search broken when HTML saved with different filename

cboos:
## Summary

Based on #106 by @moshethebeadle — replaces URL-based `isIndexPage` detection with DOM-based detection (`document.querySelector('.project-list')`), so search works regardless of how the HTML file is saved/named.

Additionally fixes the issue identified during review: the transcript filter observer (`MutationObserver`) was registered at parse time when `isIndexPage` was still `null`, causing it to run on index pages too. The observer setup is now extracted into `setupTranscriptFilterObserver()` and called from `initSearch()` after `isIndexPage` is properly determined.

## Changes

1. **From #106**: Replace `window.location.pathname` heuristics with `document.querySelector('.project-list') !== null` (detected at `initSearch()` time)
2. **New**: Extract transcript filter observer into `setupTranscriptFilterObserver()` function, called from `initSearch()` only for non-index pages
3. Updated snapshots

## Test plan

- [ ] Generate an index page and verify search works normally
- [ ] Save the index page with a different filename and verify search still works
- [ ] Verify search still works on transcript pages (non-index pages)
- [ ] Verify filter observer doesn't register on index pages

Closes #106

## Summary by CodeRabbit

* **Refactor**
  * Improved search page detection mechanism to use DOM-based identification instead of URL pattern matching.
  * Enhanced transcript filter handling with better observer setup and index management for improved search accuracy on transcript pages.

### DIFF
--- a/claude_code_log/html/templates/components/search.html
+++ b/claude_code_log/html/templates/components/search.html
@@ -88,6 +88,11 @@
 
         // Initialize UI state
         updateSearchInputState();
+
+        // Re-index when filters change (transcript pages only)
+        if (!searchState.isIndexPage) {
+            setupTranscriptFilterObserver();
+        }
     }
 
     // Build search index from current page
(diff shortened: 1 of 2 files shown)


cboos:
Oops, this one is redundant; I didn't know I could also push on the moshethebeadle:fix/isIndexPage-detection remote... So, let's keep that in #106.

PR #112, opened 2026-04-17
Ignore 'last-prompt' message type

jakeefr:
Fixes #102. Added `last-prompt` to the silent-skip list in `converter.py` alongside `file-history-snapshot` and `progress`, so Claude Code's new trailing `{type:last-prompt,...}` line no longer triggers a 'not a recognised message type' warning.

Also maintain a small open source tool in the Claude Code space, happy to contribute more if useful.

## Summary by CodeRabbit

* **Chores**
  * Improved transcript import: certain internal metadata entries are now silently ignored and unknown entry types emit a warning during import unless run in silent mode.
* **Tests**
  * Added regression tests covering silent skips, passthrough entries, and warning behavior for unknown transcript entry types.

Issue #94, opened 2026-02-16
Support messages for setting session title, name, and color


cboos:
```
Line 1 of .../067ee208-cb06-4e58-b48c-228e1f2ff87d.jsonl is not a recognised message type: {"type":"custom-title","customTitle":"Main","sessionId":"067ee208-cb06-4e58-b48c-228e1f2ff87d"}
Line 2 of .../067ee208-cb06-4e58-b48c-228e1f2ff87d.jsonl is not a recognised message type: {"type":"agent-name","agentName":"Main","sessionId":"067ee208-cb06-4e58-b48c-228e1f2ff87d"}
Line 8 of .../067ee208-cb06-4e58-b48c-228e1f2ff87d.jsonl is not a recognised message type: {"type":"agent-color","agentColor":"orange","sessionId":"067ee208-cb06-4e58-b48c-228e1f2ff87d"}
```

And `s/recognised/recognized/` ;-)

cboos:
One more, just seen now:

```
5ec93f15-d651-432a-ab1b-a96171669461.jsonl: unrecognized message type 'pr-link' - skipping
```

```
{"type":"pr-link","sessionId":"5ec93f15-d651-432a-ab1b-a96171669461","prNumber":96,"prUrl":"https://github.com/daaain/claude-code-log/pull/96","prRepository":"daaain/claude-code-log","timestamp":"2026-04-19T18:30:11.705Z"}
```

See [work/session-state-propagation.md](https://github.com/daaain/claude-code-log/blob/main/work/session-state-propagation.md).

lifeinchords:
Seeing another unrecognized message type on 1.2.0, in the same vein: `pr-link`. Claude Code emits this whenever a PR is referenced or created during the session. A single archive run hit 110 of these warnings from heavy `gh pr *` activity, so this adds up fast in PR-heavy sessions.

Structure (sanitized):

```json
{"type":"pr-link","sessionId":"<truncated>","prNumber":3,"prUrl":"https://github.com/<truncated>/<truncated>/pull/3","prRepository":"<truncated>/<truncated>","timestamp":"2026-04-19T17:19:54.494Z"}
```

Same fix pattern as the three in the OP: add a parser + renderer. A linked badge like `PR #N → url` would match how the `gh` CLI already refers to PRs.

cboos:
From my todo list #73: 
- Show transitions for cwd (now in ... directory) and gitBranch (now on ... branch)

[...]

PR #113, opened 2026-04-18
Handle custom-title, agent-name, and agent-color transcript entry types

fuleinist:
Addresses one half of #94: makes the transcript loader stop warning
on four session-metadata types that Claude Code writes as
positional markers (no `uuid`/`parentUuid`/`timestamp`):

- `permission-mode` (`{permissionMode}`)
- `custom-title` (`{customTitle}`)
- `agent-name` (`{agentName}`)
- `agent-color` (`{agentColor}`)

## Changes

- Extend `SILENT_SKIP_TYPES` (introduced in #112) with the four types
- Regression tests covering each type in `test_silent_skip.py`
- Fix `test_multiple_passthrough_types` which had `"permission-mode"`
  hard-coded as a synthetic Passthrough label
- Add `work/session-state-propagation.md` — design for the deferred
  half of #94 (propagate agent name/color onto conversational
  message titles)

## Not in this PR

Refs #94 but does **not** close it — the state-propagation
implementation lives in the linked plan doc and remains open work.

## Note on the original approach

fuleinist's original commit registered these types as
`PassthroughTranscriptEntry` and relaxed the model to make `uuid`
and `timestamp` optional. Maintainer feedback: Passthrough's role is
DAG continuity through entries that have `uuid`/`parentUuid`; these
four types lack both and shouldn't be Passthrough. This PR preserves
fuleinist's commit as an empty attribution commit and reimplements
the intent via the silent-skip list.

## Summary by CodeRabbit

* **Bug Fixes**
  * Transcript parsing now silently ignores several additional internal session metadata snapshot types and corrects a warning spelling to "unrecognized", preventing noisy or misclassified entries.

* **Tests**
  * Added and updated tests to assert the new silent-skip behavior for these metadata types and passthrough handling.

* **Documentation**
  * Added a design spec outlining session state propagation and UI decoration plans.

### DIFF
--- a/claude_code_log/converter.py
+++ b/claude_code_log/converter.py
@@ -55,6 +55,14 @@ SILENT_SKIP_TYPES: frozenset[str] = frozenset(
     {
         "file-history-snapshot",  # Internal file backup metadata
         "last-prompt",  # Trailing marker written as the last line of a .jsonl
+        # Session metadata snapshots (positional state, no uuid/timestamp).
+        # Recorded whenever Claude Code writes a state checkpoint to the
+        # transcript; see #94 for the wider "propagate this state to
+        # surrounding messages" follow-up.
+        "permission-mode",  # {permissionMode: 'acceptEdits'|...}
+        "custom-title",  # {customTitle: <str>}
+        "agent-name",  # {agentName: <str>}
+        "agent-color",  # {agentColor: <str>}
     }
 )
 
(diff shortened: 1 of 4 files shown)


cboos:
Thanks! That's a start, I'll rebase and see if I can enhance that a bit. One idea from #94 was that, at least the session name if not the session color, could be reused in the rendering somewhere.

Issue #26, opened 2025-09-16
Feature request: have claude to collect all stated requirements into a formal requirements.md file


jerzydziewierz:
For the purpose of documenting the project done,

I'd love to get a text or markdown transcript output, 

So that I can instruct `claude code` or another agent to create a definitive bullet point or checklist of features and requirements requested during the development session

note that this should downplay or plain ignore any output from the `claude code` -- as that is already present in the generated application anyway. It should only (or primarily) consider the human input. 

Note that this should not be "summarisation" but rather, detail filtering extraction and clean-up. No requirement detail should ever be lost in this process.

This could come as 

(a) just produce a markdown transcript into the specified local file, so that someone else can load it into his agent and reprocess as needed; 

or

(b) produce the transcript and add a generic command for an agent to produce the requirements.md file in one go; in such case, it is important to also handle the incremental updates over time

daaain:
Hey, I have thought about adding a Markdown renderer, the project is structured in a way that each part of the processing is separated, so it shouldn't be too hard to do the basic, full output by adding another `renderer.py` (could probably reuse some of the output format independent code though).

Filtering to user messages only currently is only done in Javascript, but I guess it would be nice to be able to do it before rendering with a CLI argument if someone never really wants to read system messages or such and have smaller output files.

I'm on the fence about having a summarisation feature in this library, so far it's only straightforward deterministic processing. But it does sound interesting, maybe we could use `claude` CLI (so no need for API integration and key management) and keep some metadata in YAML front matter in the summary Markdown to do incremental updates?

Do you feel like having a crack at any of these?

cboos:
The new `--detail minimal --format md --compact` flag combination (1.2, unreleased) will get you pretty close.

Although seeing this, I wonder if I couldn't push the cursor one notch further and only have the user messages out... (`--detail user-only`?)

cboos:
> Although seeing this, I wonder if I couldn't push the cursor one notch further and only have the user messages out... (`--detail user-only`?)

We do have that, now (already in 1.3.0).

PR #118, opened 2026-04-19
Add --detail user-only level

cboos:
Fifth detail level inspired by #26 ("have claude to collect all stated requirements into a formal requirements.md file"): a terser output than `--detail minimal` that keeps **only** content the user authored, so the transcript can be fed directly to a downstream agent that extracts requirements, builds a changelog-by-intent, etc.

## Survivors at `--detail user-only`

- `UserTextMessage` — real user prompts
- `UserSteeringMessage` — `queue-operation 'remove'` entries whose payload carries user precisions ("actually, wait — use Postgres not MySQL")
- `SessionHeaderMessage` — orientation markers so per-session boundaries remain visible

Everything else drops via `_USER_ONLY_EXCLUDE_CLASSES = (*_MINIMAL_EXCLUDE_CLASSES, AssistantTextMessage)`: assistant text, thinking, tool_use/tool_result, bash input/output, slash commands, command output, compacted summaries, memory, hooks, system.

## Level ordering

As documented in `DetailLevel`: `full > high > low > minimal > user-only` (smallest output last). Steering is user-authored content and now survives at every non-FULL level — each level keeps at least what the smaller level keeps.

## Output filename

Automatic — `variant_suffix()` already concatenates the enum value, so the new variant lands as `*.user-only.html` / `*.user-only.md`. Covered by `TestVariantSuffix`.

## UserSteeringMessage retention across levels (hierarchy fix)

`_filter_by_detail` used to drop every `QueueOperationTranscriptEntry` before the post-render filter got to look at it, silently removing `UserSteeringMessage` from HIGH/LOW/MINIMAL output. The post-render exclude chains (`_HIGH/_LOW/_MINIMAL/_USER_ONLY_EXCLUDE_CLASSES`) never listed `UserSteeringMessage`, so this was the only thing filtering it out.

Fix: `_filter_by_detail` now allows `QueueOperationTranscriptEntry` at every non-FULL level unconditionally. Steering is user-authored content — every view that claims to preserve the user's side of the conversation keeps it.

`TestSteeringHierarchy` is parametrised over HIGH/LOW/MINIMAL/USER_ONLY to lock the behaviour.

## String-content queue-operation fix

`QueueOperationTranscriptEntry.content` is `Optional[Union[list[ContentItem], str]]`. Two pre-render sites (`_filter_messages` line 1904, session-info collection line 2161) were coercing non-list content to `[]`, silently dropping steering that arrives as a plain string.

Fix: new `_queue_op_content_as_list` helper wraps non-empty strings in a single `TextContent` so downstream filters see a uniform list. `TestSteeringStringContent` covers the regression.

## Index enumeration fix (regex)

`VARIANT_ENTRY_RE` in `utils.py` used `\.[a-z]+` per suffix segment, rejecting the hyphen in `user-only`. The `.user-only.html` file was written to disk and picked up by `_enumerate_project_variants`' glob, but then silently dropped by the regex.

Fix: `[a-z]+` → `[a-z-]+` (hyphen at end of class, literal). Covered by `TestVariantEntryRegex::test_accepts_entry_points` (two new param cases) and `TestEnumerateProjectVariants::test_lists_all_variants_default_first` (now plants a user-only fixture file).

## UI label polish

`_variant_label_from_suffix(".user-only")` now renders as `"User only"` (space) rather than `"User-only"` (hyphen). The hyphenated form is kept for the CLI flag, enum value, and on-disk filename — only the human-facing label is polished.

## Tests

- `TestUserOnlyTemplateMessages` (6 tests): assistant-text drop, steering retention, tools/thinking/bash/slash drop, sidechain drop, session-header preservation, size-vs-MINIMAL invariant.
- `TestUserOnlyCli`: CLI accepts `--detail user-only` and emits `*.user-only.html`.
- `TestMinimalTemplateMessages::test_minimal_keeps_user_steering`: MINIMAL path for steering retention.
- `TestSteeringHierarchy` (parametrised over HIGH/LOW/MINIMAL/USER_ONLY): hierarchy consistency.
- `TestSteeringStringContent`: string-content queue-operation produces a UserSteeringMessage.
- `TestVariantSuffix`: USER_ONLY suffix correct for html + md.
- `TestVariantEntryRegex::test_accepts_entry_points`: hyphenated suffix cases.
- `TestEnumerateProjectVariants::test_lists_all_variants_default_first`: user-only fixture + rendered-label assertion.

Full `just test`-equivalent run (skipping pre-existing pagination SQLite failures in my local env): **929 passed, 7 skipped**. `ruff format` + `ruff check` + `pyright` clean on modified files.

Refs #26.

## Summary by CodeRabbit

* **New Features**
  * Added a new CLI detail level: --detail user-only to emit only user prompts and steering content (lowest verbosity).

* **Documentation**
  * README updated to document the new user-only detail level and its position in the verbosity ordering.

* **Output**
  * Generated output now includes a .user-only variant suffix and displays the hyphenated label as "User only".

* **Tests**
  * Expanded test coverage for the new detail level and output filename/variant behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ uvx claude-code-log@latest --open-browser
 - **Rich Message Types**: Support for user/assistant messages, tool use/results, thinking content, images
 - **System Command Visibility**: Show system commands (like `init`) in expandable details with structured parsing
 - **Markdown Rendering**: Server-side markdown rendering with syntax highlighting using mistune
-- **Detail Levels & Compact Mode**: `--detail full|high|low|minimal` filters by verbosity and `--compact` merges repeated section headings — pairs well with `--format md` to feed past conversations back to an LLM for analysis or experience building
+- **Detail Levels & Compact Mode**: `--detail full|high|low|minimal|user-only` filters by verbosity and `--compact` merges repeated section headings — pairs well with `--format md` to feed past conversations back to an LLM for analysis or experience building
 - **Floating Navigation**: Always-available back-to-top button and filter controls
 - **CLI Interface**: Simple command-line tool using Click
 
@@ -156,6 +156,7 @@ claude-code-log /path/to/project --detail low --format md --compact
 
 `--detail` levels (smallest → largest output):
 
+- `user-only` — just user prompts and steering (useful as input to a downstream agent, e.g. building a requirements doc)
 - `minimal` — user + assistant text only
 - `low` — interaction-focused; keeps WebSearch, WebFetch, and Task (agent delegations) as key signals
 - `high` — detailed but cleaned; drops system/hook noise
--- a/claude_code_log/cli.py
+++ b/claude_code_log/cli.py
@@ -515,14 +515,18 @@ def _clear_output_files(input_path: Path, all_projects: bool, file_ext: str) -> 
 )
 @click.option(
     "--detail",
-    type=click.Choice(["full", "high", "low", "minimal"], case_sensitive=False),
+    type=click.Choice(
+        ["full", "high", "low", "minimal", "user-only"], case_sensitive=False
+    ),
     default="full",
     help=(
         "Detail level for output. "
         "full: everything; "
         "high: detailed but cleaned (no system/hook noise); "
         "low: interaction-focused + key signals; "
-        "minimal: user + assistant messages only."
+        "minimal: user + assistant messages only; "
+        "user-only: only user prompts and steering (for feeding to "
+        "downstream agents, e.g. building a requirements doc)."
     ),
 )
 @click.option(
--- a/claude_code_log/converter.py
+++ b/claude_code_log/converter.py
@@ -901,7 +901,9 @@ def _variant_label_from_suffix(suffix: str) -> str:
         return "Full"
     parts = [p for p in suffix.split(".") if p]
     # Capitalise each segment; "compact" stays lowercased as the adverb.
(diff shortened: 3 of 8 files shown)


cboos:
Thanks for the two follow-up findings. Acknowledging both, declining both, reasoning below:

### 1. "Prevent session summaries from leaking into `user-only` output"

Summaries are session *labels* (Claude-authored session titles like "Session about X"), attached to `SessionHeaderMessage` for navigation — they aren't prompt-like text that a downstream agent would mistake for user input. For the #26 use-case (feeding the transcript to an agent building `requirements.md`), session labels are useful scaffolding that tells the agent when topic boundaries occur. Stripping them would cost real context for a theoretical purity gain. Keeping them.

### 2. "Treat steering messages as user-authored session starters"

A session opening with only steering content can't exist in practice: steering (`queue-operation "remove"`) is an out-of-band user message sent *while the agent is busy processing something else*. That "something else" is by definition the first `UserTextMessage` of the session. So the first entry is always a real user prompt, and steering only ever arrives after.

Even in the impossible case, the degradation is "session preview is empty" — the session header still renders, navigation still works. A graceful no-op, not a bug.

Not worth the plumbing to patch.

@coderabbitai resolve

daaain (approved):
Looks good, both the implementation and the mode, I was thinking myself that extracting just my messages and processing them could be useful to see how to improve documentation so I can steer less later!

PR #131, opened 2026-04-26
Robust within-session fork rendering: collapse parallel-tool_use forks, consistent labels

cboos:
## Summary

Long-running team-spawn / parallel-tool_use sessions surfaced a cascade
of within-session-fork rendering artifacts:

- Spurious `Fork point (1 branches)` ghost entries in the index, deeply
  nested in the parallel-tool_use case where every parallel turn became
  a 2-child DAG fork (one continuation through a `progress` passthrough,
  one dead-end `tool_result`).
- Branches labelled inconsistently across the session index, the body
  branch header, and the in-flow fork-point box — same fork shown three
  different ways (rich preview here, UUID-only there, "Branch <8 hex>"
  in the third spot).
- `from ⑂ Fork point` backlinks pointing at the wrong message after the
  Skill-pair reindex shifted indices: the cached `parent_message_index`
  on `SessionHeaderMessage` instances created earlier in
  `_render_messages` was never remapped, so the link jumped to whatever
  message ended up at the *old* index after the shift.
- Noisy `Session …: 15 roots found (7 unexpected)` warnings on
  every long compacted session, with 6 individual `Orphan node …`
  log lines per `/compact` boundary.

This PR addresses all four under one consistent framing.

## Two commits

1. **DAG: collapse parallel-tool_use forks via passthrough chain**
   - New Variant 3 in `_walk_session_with_forks`: at a fork where exactly
     one passthrough child has a live (non-structural) subtree and every
     other child is structural, follow the passthrough as the chain
     continuation and stitch the user/assistant siblings inline as
     dead-end side entries. Distinct from real user rewinds, which never
     include passthrough children.
   - Drop the `max_depth=20` clamp on `_is_structural_subtree` —
     `progress` passthrough chains under a parallel-tool_use anchor
     regularly exceed the limit, which was misclassifying them as live.
     The `seen` set + `session_uuids` filter still bound termination.
   - Classify `progress` passthroughs as expected roots
     (`_EXPECTED_ROOT_PASSTHROUGH_TYPES = {"progress"}`). Async hooks
     (`SessionStart:*`, `PostToolUse:*`) routinely fire as parentless
     roots — either as a session's first event before any user turn, or
     orphan-promoted when their spawning `tool_use` is in the discarded
     pre-compaction context. Per-orphan log demoted to DEBUG when the
     orphan is a progress passthrough; multi-root warning stays quiet.
   - `dev-docs/dag.md` gets a Variant 3 section (mermaid + detection
     criteria + summary-table row) and a new "Expected Root Types"
     subsection tabulating all six legitimate root shapes (compact
     boundary, local_command, two flavors of progress passthrough,
     session's actual first entry, subagent root).

2. **Renderer: consistent branch labels + fix stale fork-point backlinks**
   - `_reindex_filtered_context` now also remaps `parent_message_index`
     on every branch / child `SessionHeaderMessage`, not just
     `ctx.session_first_message`. Otherwise the `_pair_skill_tool_uses`
     reindex left every branch's "from ⑂ Fork point" link pointing at a
     stale index.
   - New helpers `_branch_short_uuid` and `_branch_label` are the single
     source of truth for the `Branch • <uuid8> • <preview>` (or
     `Branch • <uuid8>` when no preview) shape. Used by:
     - the body branch header title in `_render_messages`,
     - the index nav `first_user_message` in `prepare_session_navigation`,
     - the fork-point box per-branch link composition.
   - New `_enrich_branch_titles` post-pass scans the final `ctx` for
     each branch's first `UserTextMessage` when the branch's literal
     first transcript entry is an assistant (e.g. "No response
     requested." after `/exit`) or a tool_result — so the body header
     surfaces the same rich preview the index already had.
   - Fork-point indicator's "≥ 2 non-empty previews" elision relaxed to
     "≥ 2 navigable branches": the DAG-level Variant 3 already collapses
     spurious single-branch shells, so this layer surfaces the indicator
     reliably for every real fork.

## Test plan

- [x] New `TestParallelToolUseViaPassthrough` (basic shape + the >20-deep
      passthrough chain that exercises the depth-cap removal)
- [x] `TestRootClassification::test_progress_passthrough_roots_are_expected`
      covering both `parentUuid:null` and orphan-promoted shapes
- [x] `TestReindexBranchBackrefs` (`test_skill_pairing.py`) — two cases:
      unshifted (drop later messages) and shifted (drop an earlier
      message moves the fork-point's index, the backlink must follow)
- [x] `just ci` clean: 1206 tests, ruff, pyright, ty all green
- [x] Real-data verification on a 9-`/compact` 14k-node session with
      multiple parallel-tool_use turns: 5 fork points / 10 branches →
      2 fork points / 4 branches (the two remaining are real
      slash-command rewinds), and zero DAG warnings emitted.

## Summary by CodeRabbit

* **New Features**
  * Branch-aware session anchors, richer branch headers, and TOC/navigation linking to correct branch headings; branch preview text preserved for labeling.

* **Bug Fixes**
  * Better structural-subtree detection to avoid spurious forks on long passthrough chains.
  * Expected passthrough roots (progress) no longer reported as unexpected.
  * Backlink/index remapping fixed when context is filtered.

* **Documentation**
  * Expanded stitching-heuristics and expected-root taxonomy.

* **Tests**
  * New regression and robustness tests covering these behaviors.

### DIFF
--- a/claude_code_log/dag.py
+++ b/claude_code_log/dag.py
@@ -165,12 +165,30 @@ def build_dag(
                 parent.children_uuids.append(node.uuid)
             else:
                 if node.parent_uuid not in _sidechain_uuids:
-                    logger.warning(
-                        "Orphan node %s: parentUuid %s not found in loaded"
-                        " data (promoting to root)",
-                        node.uuid,
-                        node.parent_uuid,
-                    )
+                    # Progress passthroughs are async hooks: their parent
+                    # tool_use is routinely lost to ``/compact`` (the
+                    # spawning turn is in the discarded pre-compaction
+                    # context). Log at debug — the multi-root warning
+                    # already classifies them as expected roots; per-node
+                    # warnings here would just multiply the noise on long
+                    # compacted sessions. See ``_is_expected_root_type``.
+                    if (
+                        isinstance(node.entry, PassthroughTranscriptEntry)
+                        and node.entry.type in _EXPECTED_ROOT_PASSTHROUGH_TYPES
+                    ):
+                        logger.debug(
+                            "Orphan progress hook %s: parentUuid %s not "
+                            "found in loaded data (promoting to root)",
+                            node.uuid,
+                            node.parent_uuid,
+                        )
+                    else:
+                        logger.warning(
+                            "Orphan node %s: parentUuid %s not found in "
+                            "loaded data (promoting to root)",
+                            node.uuid,
+                            node.parent_uuid,
+                        )
                 # Clear the dangling parent so this node becomes a root
                 # and can participate in DAG walks
                 node.parent_uuid = None
@@ -294,7 +312,6 @@ def _is_structural_subtree(
     uuid: str,
     session_uuids: set[str],
     nodes: dict[str, MessageNode],
-    max_depth: int = 20,
 ) -> bool:
     """Check if the subtree below `uuid` contains only structural entries.
 
@@ -308,21 +325,25 @@ def _is_structural_subtree(
     The root itself is not inspected — only its descendants — because this
     check is used to decide whether a child of a fork point represents
     continuing conversation.
+
+    Traversal is bounded by ``session_uuids`` and the ``seen`` set, so a
+    long passthrough chain still terminates without a depth cap. Earlier
+    versions clamped at depth=20 and returned False for deeper chains;
+    that misclassified pure-passthrough tails (e.g. >20 chained
+    ``progress`` callbacks under a parallel-tool_use anchor) as live and
+    suppressed the spurious-fork collapse.
     """
-    stack: list[tuple[str, int]] = [(c, 1) for c in nodes[uuid].children_uuids]
+    stack: list[str] = list(nodes[uuid].children_uuids)
     seen: set[str] = set()
     while stack:
-        current, depth = stack.pop()
+        current = stack.pop()
         if current in seen or current not in session_uuids:
             continue
         seen.add(current)
         entry = nodes[current].entry
         if isinstance(entry, (UserTranscriptEntry, AssistantTranscriptEntry)):
             return False  # Found conversational content
-        if depth >= max_depth:
-            return False  # Too deep to tell — be conservative
-        for c in nodes[current].children_uuids:
-            stack.append((c, depth + 1))
+        stack.extend(nodes[current].children_uuids)
     return True
 
 
@@ -332,11 +353,22 @@ def _is_structural_subtree(
 # noteworthy.
 _EXPECTED_ROOT_SYSTEM_SUBTYPES = frozenset({"compact_boundary", "local_command"})
 
+# Passthrough subtypes that legitimately appear as roots:
+#   - ``progress`` — hook callbacks (``SessionStart``, ``PostToolUse``,
+#     ``UserPromptSubmit``, …) are async by nature. The first hook of a
+#     session has no preceding turn (parentUuid:null naturally). Hooks
+#     still in flight when ``/compact`` fires lose their spawning
+#     tool_use to the discarded pre-compaction context, so ``build_dag``
+#     promotes them to root. Both shapes are routine, not noteworthy.
+_EXPECTED_ROOT_PASSTHROUGH_TYPES = frozenset({"progress"})
+
 
 def _is_expected_root_type(entry: TranscriptEntry) -> bool:
     """Whether a multi-root entry is one of Claude Code's expected patterns."""
     if isinstance(entry, SystemTranscriptEntry):
         return entry.subtype in _EXPECTED_ROOT_SYSTEM_SUBTYPES
+    if isinstance(entry, PassthroughTranscriptEntry):
+        return entry.type in _EXPECTED_ROOT_PASSTHROUGH_TYPES
     return False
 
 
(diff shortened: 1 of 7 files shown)


cboos:
Addressed all 8 nitpicks in fixups (autosquashed onto the existing two commits, no new commits):

**Substantive**

- `renderer.py` brittle title round-trip → added `SessionHeaderMessage.preview: Optional[str]` so the fork-point box, session/graph index nav and `_enrich_branch_titles` all read the raw preview directly instead of `title.split(" • ", 2)`-decoding the composed label. Eliminates the encode/decode dance.
- `_enrich_branch_titles` length heuristic → replaced with a UUID-fallback gate: enrich only when the existing preview is empty (or, equivalently, when the branch title is the bare `Branch • <uuid8>` shape). Short slash-command captures like `"/exit"` are no longer at risk of being clobbered by a longer-but-less-informative scan result.
- Markdown branch parity → `title_SessionHeaderMessage` now surfaces `content.title` (the `Branch • <uuid8> • <preview>` already laid out by `_branch_label`) when `is_branch` is true, with a 🌿 prefix to distinguish it from trunk session headers. Closes the TODO; opted for the in-PR fix rather than a follow-up issue since the new `preview` field made it a 3-line change.

**Doc / test polish**

- `dev-docs/dag.md` taxonomy table — narrowed row 1 to "user prompt only", let row 4 own the SessionStart-hook case.
- `test/test_dag.py::test_deep_passthrough_chain_classified_structural` — docstring now explicitly states it exercises the depth-unboundedness invariant Variant 3 *depends on* (live continuation flows through user child `u2`, the Shape B collapse), not Variant 3 itself.
- `test/test_skill_pairing.py::test_parent_message_index_remapped` — comment now says "no-shift scenario; the shifting case is `…_when_fork_shifts`."
- `_branch_label` truncation: swapped `"..."` → `"…"` so the implementation matches the docstring's "80 chars + ellipsis" cap (was producing 83 visible chars).

**Skipped**

- `dag.py` Variant 3 observability `logger.debug` for non-`progress` passthroughs and `_is_structural_subtree` cache — acknowledged as non-blocking; the predicate-breadth comment already documents the trade-off and the cache is "not a hot-path concern" per the review.

`just ci` clean (1206 tests, ruff, pyright, ty all green). Branch is rebased on `main` and force-pushed.

cboos:
Second pass — addressed all 4 follow-up concerns. Squashed into the existing renderer commit.

**Major (markdown duplicate anchors)**
`format_SessionHeaderMessage` was emitting `<a id="session-{trunk_uuid8}">` for every branch — every branch's session_id starts with the trunk's uuid, so all branches under one trunk produced the same anchor and TOC links could only land on the first heading. Added `_session_anchor()` helper that returns `branch-<branch_uuid8>` for branch sids (using the deepest `@`-segment, mirroring the visible `Branch • <uuid8>` label) and `session-<trunk_uuid8>` for trunks. Both `format_SessionHeaderMessage` and `_generate_toc` route through it. The TOC now also surfaces branches with their full `Branch • <uuid8> • <preview>` label (lifted from `first_user_message` populated by `prepare_session_navigation`) and skips fork-point items (Markdown has no per-message anchors for them).

**Minor (sidechain pickup in `_enrich_branch_titles`)**
Added `if msg.is_sidechain: continue` to the inner loop. Sub-agent wrapped user prompts inherit `render_session_id = agent_parent_session` (which can be the branch sid), so without the guard they could be lifted as the branch's preview ahead of any branch-local human turn.

**Outside-diff Potential issue (junction_forward_links stale after re-reindex)**
`junction_forward_links` is populated before the optional detail-level filter calls `_reindex_filtered_context` a second time. Extended the reindex pass to remap each tuple's `branch_idx` through `index_remap`, drop tuples whose target was filtered out, and mirror the existing "≥ 2 navigable branches" elision when fewer than 2 remain. Same shape as the `parent_message_index` fix that landed in the original PR.

**Nitpick (hand-strip prefix is fragile)**
Factored `_branch_label_suffix(branch_sid, preview)` out of `_branch_label`. The fork-point box now calls the suffix helper directly instead of `_branch_label(...)[len("Branch • "):]`. Single source of truth for the format, no parsing coupling.

`just ci` clean (1206 tests, ruff, pyright, ty all green). Branch force-pushed.

cboos:
Third pass — addressed all three concerns from the latest review (1 actionable + 2 nitpicks). Squashed into the existing renderer commit; force-pushed.

**Actionable: defensive branch heading when `content.title` is empty**
`title_SessionHeaderMessage` previously matched `if content.is_branch and content.title:` — when a branch had `is_branch=True` but somehow ended up with an empty `title`, it fell through to the trunk-style path which uses `content.session_id[:8]` (i.e. the trunk's first 8 chars), giving a duplicate-anchor / confusable-heading shape. The check is now `if content.is_branch:` and the empty-title branch composes a fallback `🌿 Branch • <uuid8>` from the deepest `@`-segment — same identifier `_session_anchor` uses for the markdown anchor, so heading and anchor stay aligned. (`_render_messages` always populates `content.title` via `_branch_label` today, so this is purely defensive — but if the invariant ever drifts the heading still surfaces the branch identity instead of impersonating the trunk.)

**Nit 1: TOC fallback format mismatch**
The `_generate_toc` branch fallback was emitting ``Branch `<uuid8>` `` (backticks) while the body header uses `Branch • <uuid8> • <preview>` (bullets). Aligned the fallback to `Branch • <uuid8>` to match `_branch_label`'s shape. Same caveat — `prepare_session_navigation` always populates `first_user_message` for branches, so the fallback is defensive.

**Nit 2: tighten `_session_anchor` parameter type**
`session_or_id: Any` → `session_or_id: "SessionHeaderMessage | str"`. The duck-typed contract is now expressed in the signature; the docstring is correspondingly trimmed. (`Any` is still used elsewhere in the file for content/data shapes that genuinely can be heterogeneous.)

`just ci` clean (1206 tests, ruff, pyright, ty).

cboos:
Promoted `_branch_short_uuid` → `branch_short_uuid` in `claude_code_log/renderer.py` (drops the leading underscore so the cross-module use respects Python's loose convention) and routed all three duplicated `split("@")[-1][:8]` sites in `markdown/renderer.py` (`_session_anchor`, the defensive empty-title branch heading, the `_generate_toc` branch fallback) through the now-public helper. Single source of truth — if the suffix length or splitting rule ever changes, no risk of one site drifting from another.

Updated docstring on `branch_short_uuid` to acknowledge it's a cross-module helper.

`just ci` clean. Force-pushed (`bcf658e`).

cboos:
Added a new commit on top (953fc55) for the Markdown compact-point duplication you flagged:

The compact-point nav items in `prepare_session_navigation` carry `id = "compact-{message_index}"`. When the Markdown TOC routed those through `_session_anchor`, the `[:8]` slice of every `compact-{N}` collapsed to the literal `"compact-"` — so a long compacted session emitted multiple TOC links all pointing at the same malformed `#session-compact-` anchor target (one per compact, all duplicated, none resolving). The d4bea21c fixture has 9 `/compact` runs and was the trigger.

Skipped them in `_generate_toc` alongside the existing fork-point filter — same rationale (both navigate via `msg-d-{N}` HTML anchors that Markdown doesn't have). Markdown TOC now only surfaces session-level entries (trunk + branch headers), all with real anchor targets emitted by `format_SessionHeaderMessage`.

Branch is now 3 commits:
- 953fc55 Markdown TOC: skip compact-point nav items
- bcf658e Renderer: consistent branch labels + fix stale fork-point backlinks
- b323180 DAG: collapse parallel-tool_use forks via passthrough chain

`just ci` clean.

PR #135, opened 2026-05-05
Fix DAG cyclic-children hang and add SIGUSR1 stack dump

daaain:
Real-world transcripts can contain cyclic ``parentUuid`` chains — two entries naming each other as parent, or a self-loop. ``build_dag`` detected the cycle in the parent-walk validation step and nulled the offending ``parent_uuid``, but it had **already** populated ``children_uuids`` from those same cyclic edges. The parent chain ended up acyclic while ``children_uuids`` retained the loop.

``_walk_session_with_forks`` traverses via ``children_uuids`` with no visited-set guard, so on hitting a cyclic child it would chase the loop forever, growing ``chain``, ``queue`` and per-node state without bound. A live ``py-spy`` dump caught the hang in the wild: 50+ minutes at 100% CPU, 256 GB resident, ``chain`` filled with the same UUID thousands of times over, ``same_session_children`` containing the node itself.

## Fix

``build_dag`` is reordered into three explicit phases:

1. Clear dangling parent_uuids (orphans → roots).
2. **Break parent_uuid cycles.** Done before children population, so children can never contain a cyclic edge. Nulling the offending pointer alone is now sufficient.
3. Build parent→children links from the now-acyclic parent pointers, with self-edge and duplicate guards as belt-and-braces.

``_walk_session_with_forks`` gains a ``walk_visited`` set in its inner loop. Defence in depth — even if a future change reintroduces a cyclic child, the walk truncates the chain and warns instead of hanging.

## Regression tests

``test/test_dag.py::TestCyclicParentChain`` — five tests covering self-cycle, two-cycle and three-cycle inputs at both the structural level (``children_uuids`` acyclic) and the end-to-end level (``extract_session_dag_lines`` terminates). The end-to-end checks run under a thread-based 5 s timeout so a regression of this bug fails the test instead of hanging the entire suite. All five fail on main before the fix; all pass after.

## Diagnostic: SIGUSR1 stack dump

``cli.py`` registers ``faulthandler`` against ``SIGUSR1`` so ``kill -USR1 <pid>`` prints the live Python stack to stderr without killing the process. Unlike ``py-spy``, no root or extra install is required — handy on macOS where ``py-spy`` needs ``sudo``. The investigation that found this bug took 50 minutes; with this hook the same diagnosis takes seconds. POSIX-only; silently no-ops on Windows. Documented in CONTRIBUTING.md under "Diagnosing Hangs".

[...]

PR #138, opened 2026-05-06
Fix/prevent dag cycle

daaain:
@cboos these are addressing the warnings I mentioned in #135

I'm not super familiar with this DAG processing so please have a look!

## Make DAG cycle warnings honour silent=True

Normal generation should just output projects list

## Quiet routine multi-root warnings

The "N roots found (M unexpected)" warning fired on long sessions for
shapes that are actually routine:

* The genuine session-start user root counted as 'unexpected' alongside
  expected `compact_boundary` roots.
* Sidechain user roots with no agentId — older Claude Code Task prompts
  loaded without their trunk anchor.
* Cross-session attachment roots — `--resume` shape where the resumed
  transcript's entries reference parents that live in another loaded
  session.

Add `_classify_unexpected_roots` to filter these patterns out before
deciding whether to warn. The genuine start is identified as the
earliest non-sidechain natural root, so a sidechain orphan dated
before the real session start can't shadow it.

On the user's repower project this drops 16 multi-root warnings to 0
while keeping the legitimate orphan-node warning intact. Adds four
regression tests covering each shape, including the
sidechain-earlier-than-user-start edge case.

## Fix self-loop anchor for orphan subagent transcripts

When a sidechain transcript was loaded without its trunk session (e.g.
warmup-only subagent files), `_integrate_agent_entries` accepted the
agent's own root entry as its anchor. The subsequent re-parent step
then set `parentUuid = uuid`, producing a self-loop and a "Cycle
detected" warning at DAG construction.

Tighten the sidechain-anchor predicate to require a cross-agent
boundary (parent belongs to a different agent's transcript). Add a
real-world repro fixture and three regression tests, including one
that guards the legitimate nested A→B anchor case.

## Release Notes

* **Bug Fixes**
  * Suppressed unnecessary diagnostic warnings during silent and cache rebuild operations
  * Prevented unintended cycles in transcript chain reconstruction
  * Improved handling of orphaned transcript entries and cross-session attachments

* **Tests**
  * Enhanced test coverage for diagnostic message suppression and silent modes
  * Added regression tests for cycle prevention in session transcript integration


cboos (approved):
I had a set of sessions that, when processed together, exhibited the warnings. With these changes they are gone and the content looks good, so... approved!

PR #140, opened 2026-05-07
System info cosmetic improvements + chain-pairing fix (#137)

### LINKED ISSUE #137: System Info cosmetic improvements
I noticed the *Fixed /context dumping its rendered ASCII visualization grid into the conversation, wasting ~1.6k tokens per call* bullet point from the 2.1.129 entry in the Claude Code changelog, so I wanted to see how it affected us.

Sort of:

Image
<br>

That would look much better already with a `<pre>` in the mix:

Image
<br>

Bonus points for disentangling the `/color` and `/context` calls, seems like we have *two* pair-middle...

Image

cboos:
## Summary

Closes #137. Two related bugs in system-info rendering, surfaced by the same user transcript and fixed together because they fix the same user-facing problem (`/context` ASCII grid rendered unusably).

- **Cosmetic** (`html/system_formatters.py` + new `.system-content` CSS): multi-line system content (e.g. `/context` token-usage grid) wraps in `<pre class='system-content'>` so whitespace is preserved. CSS uses `white-space: pre` (not `pre-wrap`) so grid rows don't wrap mid-row on narrow viewports — `overflow-x: auto` provides scroll instead. Single-line system messages keep inline rendering.
- **Pairing chain** (`renderer.py::_try_pair_by_index`): chains of system messages (each `parentUuid` = previous system's `uuid`, common with `/context` / `/cost` multi-step output) used to render as `pair_first → pair_middle × (N-2) → pair_last` — fake N-tuple. Skip `_mark_pair` when the candidate parent is already paired as a child; chains break into pairs of two from the leading edge.
- **Markdown mirror** (`markdown/renderer.py::format_SystemMessage`): the same multi-line bug existed on the Markdown side (raw newlines in inline-paragraph context get collapsed). Routed through the existing `_code_fence` helper so the fence widens adaptively past any backticks in the body.

Two commits are kept self-contained rather than squashed:

- `7edf13e` — HTML `<pre>` wrap + chain-pairing fix + 6 dedicated regression tests + the snapshot updates for the new CSS rule.
- `12bf131` — Markdown mirror + 4 dedicated regression tests; no snapshot churn.

## Test plan

- [x] `just test` — 1576 pass, 7 skipped (was 1566 on `main`; +10 new tests across the two commits)
- [x] `just test-tui` — 66 pass
- [x] `just test-browser` — 39 pass, 1 skipped
- [x] `just ci` (full quality gate) — green
- [x] `ruff check` — clean
- [x] `pyright` on touched production files — clean
- [x] Snapshot updates in `test_snapshot_html.ambr` reviewed: exactly the new `.system-content` CSS rule × 7 snapshots, no content drift
- [x] New regression tests verified to fail pre-fix and pass post-fix:
  - `TestSystemMessageChainPairing` (3 tests) — 4-chain produces `[pair_first, pair_last, pair_first, pair_last]` (was `[pair_first, pair_middle, pair_middle, pair_last]`); 3-chain produces `[pair_first, pair_last, none]`; 8-chain never produces `pair_middle` (only legitimate source remains the `UserSlash → Slash → CommandOutput` triple, different code path)
  - `TestFormatSystemContent` (3 tests) — single-line stays inline, multi-line wraps in `<pre class='system-content'>`, level icon present in both
  - `TestMarkdownFormatSystemMessage` (4 tests) — mirror coverage on the Markdown side, including a backtick-adversarial test that pins the adaptive-fence contract via a regex backreference

## Summary by CodeRabbit

* **Bug Fixes**
  * System-message pairing corrected to avoid incorrect chain/sibling groupings.

* **New Features**
  * Multi-line system messages render in a monospace preformatted block with the level icon and horizontal scrolling; single-line remains inline.

* **Tests**
  * Added unit and regression tests for system-message rendering and pairing in HTML and Markdown.

* **Documentation**
  * Updated snapshot testing guidance to avoid concurrent update races.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -116,16 +116,26 @@ just test-cov
 Snapshot tests detect unintended HTML output changes using [syrupy](https://github.com/syrupy-project/syrupy):
 
 ```bash
-# Run snapshot tests
+# Run snapshot tests (parallel mode is fine for read-only runs)
 uv run pytest -n auto test/test_snapshot_html.py -v
 
 # Update snapshots after intentional HTML changes
-uv run pytest -n auto test/test_snapshot_html.py --snapshot-update
+# IMPORTANT: run --snapshot-update WITHOUT -n auto (see warning below)
+uv run pytest test/test_snapshot_html.py --snapshot-update
 ```
+
+> **Warning — don't combine `--snapshot-update` with `-n auto`.** Syrupy
+> and pytest-xdist race when writing snapshot files in parallel: the
+> `.ambr` file ends up truncated (observed: ~6000 lines silently
+> deleted on a single run, leaving the file structurally broken but
+> still passing on next read). Run `--snapshot-update` serially. This
+> is also why pytest is **not** configured with a default `-n auto`
+> in `pyproject.toml`; the `just test` recipes opt in for read-only
+> runs where the race doesn't apply.
 
 When snapshot tests fail:
 1. Review the diff to verify changes are intentional
-2. If intentional, run `--snapshot-update` to accept new output
+2. If intentional, run `--snapshot-update` (serially) to accept new output
 3. If unintentional, fix your code and re-run tests
 
 ### Test Prerequisites
--- a/claude_code_log/html/system_formatters.py
+++ b/claude_code_log/html/system_formatters.py
@@ -21,6 +21,14 @@ from ..models import (
 def format_system_content(content: SystemMessage) -> str:
     """Format a system message with level-specific icon.
 
+    Multi-line content (ASCII visualisations from ``/context``,
+    ``/cost``, etc., or any system entry whose cleaned text spans more
+    than one line) renders inside a ``<pre>`` block so whitespace and
+    grid alignment are preserved. Single-line system messages
+    (``/status`` dialog hints, ``/init`` cleaned-name forms, …) keep
+    the existing inline rendering — adding a block-level ``<pre>`` to a
+    one-liner would cost vertical space without benefit.
+
     Args:
         content: SystemMessage with level and text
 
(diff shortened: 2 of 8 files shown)


cboos:
(Claude) Status update on the latest CodeRabbit review (run `78c1a0ee-...` then re-evaluated on the rebased tip):

**Applied** — quote-flexible regex in `_system_pair_classes` ([commit 15367fa](../commit/15367fa)). The Jinja template only emits single-quoted class attrs today, but the back-referenced `(['"])...\1` form is robust to either style and the helper now also asserts at least one match before continuing, so a future template change fails the helper directly instead of silently returning an empty list.

**Declined with documentation** — `addopts = "-n auto"` in `pyproject.toml`. Setting xdist parallel mode as the project default would silently corrupt snapshot files: `pytest -n auto --snapshot-update` races on the syrupy `.ambr` file (we hit this earlier in this PR — a single parallel update silently truncated ~6000 lines of HTML snapshots). The `just test` recipes opt into `-n auto` for read-only runs where the race doesn't apply; defaulting it project-wide would mean every developer running `pytest --snapshot-update` from the project root would silently corrupt snapshots. CONTRIBUTING.md now documents this trade-off in the Snapshot Testing section so the next "why isn't this the default?" question has a reachable answer ([commit 15367fa](../commit/15367fa)).

**Declined as judgment call** — parametrizing the three chain-length tests. Each of `test_chain_of_four_pairs_into_two_doubles` / `test_chain_of_three_pairs_first_two_only` / `test_no_middle_role_in_chain` carries a focused docstring naming the narrative ("two doubles", "first two only", "no middle role"). A `pytest.mark.parametrize` would either lose those or fold them into test IDs — for three fixed cases I think the named-test version reads better, especially when one of the three uses a different assertion shape (n=8 case asserts on the `pair_middle` predicate, not the full role list).

Earlier round of findings ([commit 19be0d7](../commit/19be0d7)) — both fully applied: CSS specificity tightened to `.content pre.system-content` (the original `.system-content` was being silently overridden by the existing `.content pre` rule with `pre-wrap + break-word`, so the cosmetic fix was *almost* a no-op without this); pairing guard tightened to `parent.pair_first is None and parent.pair_last is None` (the original guard caught chains but missed sibling-overwrite — two system messages with the same `parentUuid` produced a half-pair orphan). New regression test `test_siblings_share_parent_only_first_pairs` covers the sibling case directly.

Thanks for the careful read — the CSS-specificity catch in particular would have been embarrassing to ship as a silent no-op.

PR #147, opened 2026-05-08
Render the built-in Monitor tool with Task-end backlink (#142)

cboos:
## Summary

Closes #142. Two related pieces in one commit:

- **Specialised rendering for the built-in `Monitor` tool** (the shell-stream watcher with `description` / `command` / `timeout_ms` / `persistent`). New `MonitorInput` (Pydantic) + `MonitorOutput` (dataclass), factory parser, HTML title `🔭 Monitor <description>` + 4-row params grid + collapsible command via the existing `render_collapsible_code` helper, mirrored Markdown.
- **Task-end → originating tool_use backlink.** When a `<task-notification>` carries `<tool-use-id>` (e.g. the harness emits this when a Monitor stream ends), the notification's Task ID value is wrapped in an anchor pointing at the originating tool_use card. Extended `task_notification_factory` to extract `<tool-use-id>` (older notifications without the field still parse — `tool_use_id` defaults to `None`). New renderer pass `_link_tool_use_notifications` runs after `_link_async_notifications` and sets `spawning_task_message_index` only on virgin notifications carrying a matching id. The HTML formatter now wraps the Task ID's `<code>` in `<a class='task-notification-backlink'>` whenever the index is set; the previous separate "Spawn" row is replaced by the Task-ID-as-link affordance. This unifies the async-agent and Monitor flows behind one rendering shape on the HTML side.

The Markdown side keeps the existing two-row shape — the markdown renderer's own comment notes that "Markdown only has session-level `<a id=…>` anchors", so message-level backlinks aren't structurally available in MD output.

Out of scope: `mcp__plugin_clmail_clmail__control` with `action="monitor"` (the actor-monitoring MCP tool) — distinct shape, can be a follow-up issue.

## Test plan

- [x] `just test`: 1626 pass, 7 skipped (was 1607 on `main`; +19 new tests)
- [x] `just test-tui`: 66 pass
- [x] `just test-browser`: 39 pass, 1 skipped
- [x] `just ci` end-to-end clean (ruff + pyright + ty all green)
- [x] New regression suite `test/test_monitor_rendering.py` (20 cases) covers:
  - `TestMonitorInputModel` / `TestMonitorOutputParser` — direct factory tests including the missing-task-id fallback
  - `TestMonitorHtmlFormatter` — short-command-inline, long-command-collapsible, optional-fields-omitted, output paragraph
[...]

PR #152, opened 2026-05-10
Render ScheduleWakeup and Cron* tools (#148)

### LINKED ISSUE #148: Add support for ScheduleWakeup and other Cron* tasks
Image
<br>

<details><summary>ScheduleWakeup message example</summary>

```json
{
  "parentUuid": "8a1c8fbb-b636-4c7d-a24b-953a2852619a",
  "isSidechain": false,
  "message": {
    "model": "claude-opus-4-7",
    "id": "msg_01LiEYhj6Hwg6jcoMSbdx6ni",
    "type": "message",
    "role": "assistant",
    "content": [
      {
        "type": "tool_use",
        "id": "toolu_01Q4rSkqkkKjnqbHKqP1Gt7M",
        "name": "ScheduleWakeup",
        "input": {
          "delaySeconds": 240,
          "reason": "First parent loop tick at +4min — by then (...)"
        },
        "caller": {
          "type": "direct"
        }
      }
    ],
    "stop_reason": "tool_use",
    "stop_sequence": null,
    "stop_details": null,
    "usage": {
      "input_tokens": 1,
      "cache_creation_input_tokens": 157,
      "cache_read_input_tokens": 105721,
      "...",
      "speed": "standard"
    }
  },
  "requestId": "req_011Cak6isXUTj1afT5aZFtR5",
  "type": "assistant",
  "uuid": "2514ea31-9c9f-4487-8bc1-ec8ed38b9284",
  "timestamp": "2026-05-05T21:03:53.150Z",
  "userType": "external",
  "entrypoint": "cli",
  "cwd": ".../claude-code-log/experiments/loop",
  "sessionId": "6d1760b6-5a8d-4693-b6ee-f4e27e427e88",
  "version": "2.1.128",
  "gitBranch": "HEAD"
}
```

</details>

cboos:
## Summary

Closes #148. Adds specialised rendering for the built-in scheduling tool family — `ScheduleWakeup`, `CronCreate`, `CronList`, `CronDelete` — which previously fell through to the generic params-table render. Same per-tool-family pattern as the `Monitor` tool work in #147.

Two commits, kept separate for the post-mortem story:

- **`<sha1>` — core implementation** (4-tool family rendering + 24 regression tests). Models, factory parsers (with graceful-degradation fallbacks), HTML/Markdown formatters, title methods, and a JSONL fixture covering one call per tool. Shared `_format_collapsible_prompt` helper between `ScheduleWakeup` and `CronCreate` (both carry a long prompt field that benefits from the line-count-badge collapsibility used by Monitor's `command`).
- **`<sha2>` — fix-up** (CronList HTML title duplication). `title_CronListInput` originally passed `"CronList"` as the summary to `_tool_title`, which already emits the tool name from `content.tool_name` — duplicating it in the HTML output. Fix is one line; tests tightened with a cardinality check (`html.count("CronList") == 1`) and a shape-specific check (`"<span class='tool-summary'>CronList" not in html`) to catch the same regression class via different mechanisms.

## Defensive scaffolding heads-up

`ScheduleWakeup` is the only tool with real fixtures derived from actual transcripts. `CronCreate` / `CronList` / `CronDelete` fixtures are constructed against the documented schemas — no real invocations exist in the project's transcripts yet (the tools are listed as deferred but haven't been used in the wild). The output parsers degrade gracefully: every output dataclass carries a `text` field even when structured parsing succeeds, and renderers fall back to raw text whenever parsing yields no structured fields. So when real invocations show up and the format turns out to differ from the fixture, the rendering will visibly degrade to text rather than crash. Worth a follow-up to tighten the parsers when real Cron* output lands.

## Test plan

- [x] `just test`: 1661 pass, 7 skipped (was 1635 on `main`; +26 new)
- [x] `just test-tui`: 67 pass
- [x] `just test-browser`: 39 pass, 1 skipped
- [x] `just ci`: ruff + pyright + ty all green
- [x] New regression suite `test/test_cron_rendering.py` (24 cases):
  - 4 input model tests
  - 8 output-parser tests (happy-path + graceful-degradation for each tool)
  - 6 HTML formatter unit tests (grids, collapsibility, optional-flag omission, structured-vs-fallback CronList)
  - 6 fixture-driven end-to-end tests (real renderers against `test/test_data/cron_tools.jsonl`, asserting titles + grids + structured cron list + Markdown inline-code escaping)
- [x] No snapshot drift — no existing fixtures touch these tools

## Summary by CodeRabbit

* **New Features**
  - Added ScheduleWakeup and typed cron tools (create, list, delete) with structured, user-friendly status outputs.

* **UI / Rendering**
  - Improved HTML/Markdown presentation: icon titles, collapsible prompt bodies, compact job tables, and backlinking from job IDs to their creation cards.

* **Tests**
  - Added unit and end-to-end tests with fixture coverage for scheduling and cron flows.

* **Documentation**
  - Clarified transcript icon/emoji handling and widened emoji recognition ranges.

[Review Change Stack](https://app.coderabbit.ai/change-stack/daaain/claude-code-log/pull/152)


cboos:
(Claude) Status update on the latest CodeRabbit review:

**One actionable finding applied** — `c4166ff`.

Real bug in `parse_cronlist_output`: the optional `[flags]` group at the end of `_CRON_LIST_LINE_RE` greedily captured *any* trailing bracketed text, including legitimate prompt suffixes. A row like `- cj_xyz: 0 9 * * * => /loop tick [important context]` parsed with `prompt='/loop tick'` and silently dropped the `[important context]` suffix.

Fix follows CR's suggestion: validate captured flags against a known-flags whitelist (`_KNOWN_CRON_FLAGS = {'recurring', 'durable'}`). When none of the captured tokens match a known flag, the bracketed content is restored to the prompt and the row renders as flag-less.

Two new regression tests in `TestSchedulingOutputParsers`:

- `test_cronlist_preserves_bracketed_prompt_suffix` — pins the exact CR scenario.
- `test_cronlist_keeps_real_flags_alongside_known_tokens` — confirms recognised flags still parse normally.

While I was in the parser, also folded in a stylistic micro-nit from the earlier monk review on this branch (msg #2779): `True or None` → `True if X else None` for self-documentation. Same behaviour, clearer intent.

CI on alice (pre-push): `test_cron_rendering.py` 26/26 pass; ruff + pyright clean on touched files.

Thanks — the bracketed-suffix catch is a nice one. The original parser shape was overly trusting of the `[...]` syntax matching only flag-like content.

cboos:
(Claude) Polish round on user feedback — `6f34c54`.

Three concrete changes after rendering against real-data fixtures (live `CronCreate` / `CronList` / `CronDelete` experiment in a session, captured before this commit):

**1. Wrench prefix gone from ⏰-titled tools.** The transcript template prepends `🛠️` unless the title already starts with an emoji recognised by `html/utils.py::starts_with_emoji`. That function whitelisted several Unicode ranges but missed the **Misc Technical** block (U+2300–U+23FF) where `⏰` (U+23F0) lives — so all four scheduling tools rendered as `🛠️ ⏰ <name>`. Added the missing range to the whitelist.

**2. ScheduleWakeup / CronCreate body = prompt only.** Both tools' titles already carry the most informative scalars (`+<delay>s — <reason>` for ScheduleWakeup, `<cron>` for CronCreate). The previous params-table body repeated them and added a `recurring`/`durable` row that the harness's own confirmation echoes back in human form anyway. Body is now just the collapsible prompt via the existing `_format_collapsible_prompt` helper. The unused `_row_html` helper that backed the table is removed.

**3. Parser regexes updated to match real harness output.** The live experiment showed my synthetic fixtures and parser regexes were misaligned with what the harness actually emits:

- **CronCreate**: `Scheduled cron job <id>` → real: `Scheduled recurring job <id> (Every 2 minutes). Session-only…` (or `one-shot` for non-recurring).
- **CronList row**: `- <id>: <cron> => <prompt> [flags]` → real: `<id> — <description> (recurring) [session-only]: <prompt>…`. No leading hyphen, em-dash separator, parens for kind, brackets for scope, no cron expression (only a human-readable description), prompt often truncated by the harness with `…`. Regex rewritten; `CronListItem.cron` field renamed to `description` to reflect actual content.
- **CronDelete**: `Deleted cron job <id>` → real: `Cancelled job <id>.`. Parser unchanged (still text-verbatim); fixture and tests updated.

The previous fix for the bracketed-prompt-suffix bug (CR finding earlier on this PR) and the `_KNOWN_CRON_FLAGS` whitelist are dropped — the new format separates flags (parens) from prompts (after a colon), so there's no greedy-bracket vulnerability anymore.

**Bonus: tool-renderer skill updated.** Added a "Watch out: the template's wrench-suppression is emoji-range-gated" section listing all recognised Unicode ranges and the verification recipe. Future tools using emoji from outside the whitelist will hit the same trap I did, so the skill now warns about it explicitly.

CI: 1664 unit pass, ruff + pyright clean. Test count net change reflects the rewrite (removed 2 obsolete bracket-handling tests, added 5 real-format tests including a wrench-suppression regression guard).

Fixture rendered locally to verify: `🛠️ ⏰` co-occurrence is zero across the rendered HTML.

cboos:
(Claude) Round-4 — `678b9ea`. CR's nitpick + actionable both addressed.

**Applied** the bail-out approach: `parse_cronlist_output` now returns `jobs=[]` if any non-empty line fails the row regex, triggering the renderer's existing "no jobs → raw text" fallback for the whole list. This avoids the partial-table failure mode where a 9-of-10-rows match would silently hide the 10th unparsed line behind structured chrome.

Blank lines are still skipped without bailing (carve-out for whitespace; verified by a separate test).

**Skipped** the alternative actionable suggestion (preserve unparsed rows as sentinel `CronListItem` entries with a `raw=True` flag). Bail-out is simpler — no new dataclass field, single return path — and matches the existing fallback semantics at list granularity. Either shape addresses the underlying issue.

CR's exact suggested test (`test_cronlist_mixed_lines_falls_back_to_raw_text`) lands verbatim. Added a companion `test_cronlist_blank_lines_dont_bail_out` to lock in the whitespace carve-out.

CI: 29/29 in `test_cron_rendering.py`; ruff + pyright clean.

cboos:
(Claude) Visual polish round on user feedback after rendering against the real fixture — `7f4ff89`.

**1. ScheduleWakeup / CronCreate prompts render as Markdown.** The prompt is typically Markdown content (slash commands, inline code, prose) rather than preformatted text. HTML side switches from `_format_collapsible_prompt` (which wrapped in `<pre>`) to `render_markdown_collapsible`; Markdown side stops wrapping in `_code_fence` and emits the prompt directly. The unused `_format_collapsible_prompt` helper is removed.

**2. Light borders + padded cells on CronList table.** Default unstyled tables read as raw rows pasted into the page; new CSS class `.cronlist-output-table` matches the visual weight of the sibling `.system-content` / `.command-output-content` panels (#137 lineage).

**3. Cross-link CronList row ids and CronDelete result ids back to the originating CronCreate card.** New renderer pass `_link_cron_jobs_by_id` builds a `job_id → message_index` map from CronCreate outputs and stamps `creating_call_message_index` on each consumer (`CronListItem`, `CronDeleteOutput`). HTML formatters wrap the rendered id in `<a class='cron-id-backlink'>` when set.

Same pattern as PR #147's Monitor → TaskNotification backlink, but in the reverse direction (consumer → originating call) and matched by job id rather than tool_use_id. Markdown side keeps the inline-code id without a clickable link (per the established finding from #147 — Markdown only has session-level anchors, no message-level targets).

**4. CronDelete parser extracts the cancelled job id** so the cross-link pass has a key. New `_CRON_DELETE_ID_RE` and `CronDeleteOutput.job_id` field; falls back to `None` for unrecognised formats.

The cross-link affordance is subtle (dotted underline, inherited colour, accent on hover) — visible to anyone scanning for affordances but doesn't draw attention away from the structured content.

Tests: 2 new cross-link assertions (`test_html_cronlist_id_links_back_to_croncreate`, `test_html_crondelete_result_links_back_to_croncreate`), plus updates to the prompt-rendering / formatter assertions for the new shapes. Replaced the brittle `html.count("CronList") == 1` assertion (CSS comments now legitimately mention the tool name) with the more specific `"<span class='tool-summary'>CronList" not in html` shape check — same regression class, no false positives from documentation strings.

Snapshot drift is 308 lines added across 8 fixtures — all CSS additions for the new `.cronlist-output-table` and `.cron-id-backlink` rules. No content drift.

CI: 1667 unit pass, ruff + pyright clean.

cboos:
(Claude) The latest CR review's actionable (`currentColor` → `currentcolor` in message_styles.css:485) was already applied in `9946f0e` — CR's review was generated against the previous commit (`7f4ff89`) before the fix landed. CI is now fully green on `9946f0e` (CodeRabbit + 11 test matrix). Branch ready to merge.

PR #167, opened 2026-05-23
Filter hook-injected synthetic user turns at HIGH detail and below

cboos:
## Summary

External tooling — `clmail`, `clmail-monitor`, and similar — uses Claude
Code's `UserPromptSubmit` hook to inject single-line notifications like
`[monitor] alice idle` or `[clmail] You've got a new mail (#3017)` into
the conversation. These arrive in the JSONL as full `type: user` entries
with no structural marker, so the renderer was treating them as ordinary
human prompts: one full `## 🤷 User` heading per notification. In one
real cmail session at `--detail low`, that pollution amounted to 294
synthetic user-turn headings — drowning out the actual conversation.

This PR teaches the renderer to recognise these hook-injected turns,
render them compactly at `DetailLevel.FULL`, and drop them entirely at
`HIGH` and below. The behaviour matches the pre-existing "no system /
hook noise" semantics already applied to `HookSummaryMessage` /
`HookAttachmentMessage` at HIGH — same `_HIGH_EXCLUDE_CLASSES` filter,
same gating.

### Detection

`detect_hook_notification()` in `factories/user_factory.py`:

- Regex `^\s*\[(monitor|clmail)\]\s*(.*?)\s*\Z` (sources list is a
  module-level tuple — easy to extend).
- **Single-line only.** A real user prompt that *starts* with
  `[monitor] foo\n\nActually, please continue...` is preserved intact;
  the multi-line guard rejects it from hook treatment so it renders as
  a normal user turn.
- Non-hook bracketed markers (`[Request interrupted by user]`,
  `[Image #N]`, `[@filename]`, `[from alice]`) are unaffected.

### Rendering

A new `UserHookNotificationMessage` content type wraps detected turns:

- **FULL detail**: compact inline marker. Markdown emits a single
  italic line `*[monitor] alice idle*`; HTML emits a pill with
  `class="message user hook-notification"`, dimmed border, monospace
  body. Full fidelity is preserved at FULL.
- **HIGH and below**: filtered out by `_HIGH_EXCLUDE_CLASSES` (the
  same mechanism already used for `HookSummaryMessage` and
  `HookAttachmentMessage`).

### Latent markdown-renderer fix

`MarkdownRenderer._render_message` previously only emitted content when
a title was present, silently dropping the body of any standalone
message with an empty title. The new headless-content branch (`elif
content: parts.append(content)`) only catches genuine standalone
content carriers — paired partners short-circuit earlier via
`is_middle_in_pair` / `is_last_in_pair`.

## Test plan

- [x] `uv run pytest test/test_hook_user_notifications.py -v` — 14 new
      tests covering detector, factory wrapping, and FULL/HIGH/LOW
      filtering for both Markdown and HTML.
- [x] `just ci` green (1997 tests / 0 ruff / 0 pyright / ty clean).
- [x] HTML snapshot tests updated (CSS additions for the new
      `.hook-notification` pill) — reviewed before accepting.
- [ ] Manual sanity check: re-render the cmail session that motivated
      this change and confirm the 294 synthetic headings are gone at
      LOW while real user turns survive.


cboos:
Build failures due to:
> Update - We are seeing an increased rate of authentication failures for app installation tokens, affecting approximately 1% of tokens.  We are continuing to investigate.
May 23, 2026 - 16:01 UTC

cboos:
> This PR teaches the renderer to recognise these hook-injected turns,
render them compactly at DetailLevel.FULL, and drop them entirely at
HIGH and below

While the example of what happens with this "clmail" example certainly happens with other plugins and can be generalized, hard-coding regexps for that specific plugin is certainly not the way to go. We'll need to integrate that filtering with the #166 effort.

cboos:
(Claude) Closing in favour of the plugin system landed by PR #166 and PR #169.

## Why this PR is being closed

PR #166 (plugin system RFC, merged at 9926c96) and PR #169 (implementation, awaiting merge ack) together provide a generic plugin mechanism — `MessageTransformer` + plugin-defined `MessageContent` subclasses with class-side `format_markdown` / `format_html` / `title` methods and a `detail_visibility` class attribute — that fully subsumes this PR's functionality. The upcoming `claude-code-log-clmail` plugin will ship the same `[monitor]` / `[clmail]` demotion via that contract, with the regex and detection logic moving from core to the plugin.

The framing here — hard-coding two specific external hook sources (`monitor`, `clmail`) in `claude_code_log/factories/user_factory.py` — turned out to be the wrong layer. Those sources belong with the plugin that *causes* the notifications, not in core's user-message factory. That observation from the user during review of this PR is what triggered the cross-cutting design discussion that produced PR #166's unified plugin design.

## What carries forward into the plugin

This PR's design work shaped the v1 plugin contract in concrete ways:

- The four-deliverable framing — **matcher**, **MessageContent subclass**, **format / title methods**, **detail-level filter membership** — became the structural skeleton of `MessageTransformer` plus plugin-defined subclasses in PR #166.
- The multi-line guard (real user prompts that start with `[monitor] foo\n\nActually do X` must be preserved, not demoted) is documented in PR #166's worked example as a plugin-side concern, not a core contract — the right separation.
- The CSS pill (`div.hook-notification` with dimmed border + monospace body) is reference styling future plugin authors can adopt or supersede.
- The test fixture `test/test_data/hook_user_notifications.jsonl` (9 entries: pure hook lines, multi-line guard cases, real markers like `[Image #N]` / `[Request interrupted by user]` that must be preserved) is reusable as-is in the plugin's test suite.

A future `claude-code-log-clmail` plugin author can use this PR's diff as a near-complete reference for what to ship — every piece relocates, nothing is lost.

## Migration path: full relocation, not deletion-only

An earlier point in the design discussion considered a smaller v1 scope (existing-variants-only) under which this PR's migration would have been deletion-only in core. The user's question that prompted PR #166's unified design changed that calculus: unified-v1 lets plugins own their `MessageContent` subclasses outright, which means the migration of this PR's contents is full relocation rather than deletion-only.

Concretely, the future clmail plugin will own:

- `UserHookNotificationMessage` (the class)
- `format_UserHookNotificationMessage` / `title_UserHookNotificationMessage` (as class-side `format_markdown` / `title` methods on the plugin-owned subclass)
- The CSS pill styling
- The 14 regression tests from `test/test_hook_user_notifications.py`
- The `_HIGH_EXCLUDE_CLASSES` membership (migrating to `detail_visibility = FULL` on the class)

Bigger one-shot relocation, but architecturally cleaner: core stops carrying clmail-specific assumptions, and the plugin becomes the single source of truth for both the matchers and the rendering.

## Branch retention

Branch `dev/filter-hook-turns` is left in place as a reference for the future plugin author. The fixture file and test cases are particularly reusable.

Issue #159, opened 2026-05-17
Make "--depth tool" the default detail level? (and deprecated "--detail)


cboos:
> Well, I don't have enough practical experience yet, since the feature is very new, but the more I look at the `--detail high` output, the more I think it should probably become the default... The `--detail full` would still be there when we'd want to have a look at *everything* (in which case, all the hooks, the system information notices, etc. would have to be there). 

 _Originally posted by @cboos in [#129](https://github.com/daaain/claude-code-log/issues/129#issuecomment-4320738635)_

Open for discussion... let's stay at `--detail full` as default for a few weeks or releases (starting with 1.3.0).

cboos:
Speaking of modifying the level of details... From #176:

> I keep getting confused with the "level"... I think we made the wrong call, semantically speaking. When you write `--detail high` and `--detail low`, it is clear enough. But I can't help to associate "high" with "seeing things from a high-level perspective" and "low" with "low-level details", that is, just the opposite.

As I don't want to change the meaning of `--detail` (and it wouldn't work, `--detail all` or `--detail high` clearly means high level of details), what about a symmetric option? (only one could be selected at runtime).

| Detail (`--detail`) | P.O.V (`--pov` or `--level`)  | Interpretation |
|--|--|--|
| all     | lowest | See everything we can squeeze out from a transcript |
| high | low | See mostly everything, except the really noisy stuff (hooks, system messages) |
| low | high | Focus on the interesting content |
| minimal | agent | Focus on user/agent interaction only |
| user-only | user-only | Focus on user input only (modulo the recaps, see #179). |

cboos:
`--detail` could be marked as deprecated, with a hint to prefer `--level` for its clearer semantics.

cboos:
One alternative that occurred to me is to stop framing this as a "detail level" altogether, and instead express it as a hierarchical output filter.

For example:

```text
--exclude-below none|tool|agent|assistant|user
```

where the relevant hierarchy is roughly:

```text
session > user > assistant > agent > tool > hooks
```

The semantics would be:

* `none`: keep everything
* `tool`: exclude hooks
* `agent`: exclude tools and hooks
* `assistant`: exclude agents, tools and hooks
* `user`: exclude assistant, agent, tool and hook messages

I like this because the values are no longer trying to encode an ordering themselves (`high`, `low`, `minimal`, etc.). They are simply names of nodes in the message hierarchy, while the option (`exclude-below`) expresses the operation.

[...]

Issue #111, opened 2026-04-16
Support "recap" – system subtype "away_summary"


daaain:
More info: https://www.reddit.com/r/ClaudeCode/comments/1slzgtd/analyzed_the_new_recap_feature/

Example:

> {"parentUuid":"0d4221bb-4b34-42eb-828a-8892e725be2b","isSidechain":false,"type":"system","subtype":"away_summary","content":"We're adding a project-level layout to validate projectId in the prepare route tree and redirect on invalid IDs. I just created the layout file and was about to run type-checking and test it.","timestamp":"2026-04-16T11:52:02.108Z","uuid":"e2066dc9-672a-48ff-b6c9-df29103572bf","isMeta":false,"userType":"external","entrypoint":"cli","cwd":"/app","sessionId":"4520f070-9e99-41bb-9400-2efd7eda4632","version":"2.1.110","gitBranch":"asdf","slug":"vectorized-giggling-sedgewick"}

This is particularly interesting because it summarises activity above, so we could potentially use it in some activity summary?

Though from the Reddit thread it sounds like when and how it happens depends on a few different factors, so there could be sessions without any.

In any case, at the very least we need to make sure they render nicely.

cboos:
Yes, the recap feature is quite nice: when you have many independent sessions, it's quite useful to be reminded of what was going on there. What *I* find annoying about it is the constant reminder about how to disable it ;-) We're not going to render that...

<img width="1218" height="470" alt="Image" src="https://github.com/user-attachments/assets/804a92d7-0c98-43bb-af85-e70827241730" />

Also, need to fix the levels here, the system info shouldn't nest in the hook.

daaain:
Yes! It's a bit silly that the reminder doesn't go away... Great idea to filter it out!

cboos:
Such a recap can also be useful for at least the `--detail high` level, if not the `--detail low`. We'd have to add a mechanism to not filter-out every system information message systematically (or add a "detail" field, that we can test against).

PR #176, opened 2026-05-29
Co-locate the away-summary detail rule on AwaySummaryMessage

cboos:
## What

Declares `detail_visibility: ClassVar[DetailLevel] = HIGH` on `AwaySummaryMessage` (models.py) instead of encoding its detail rule via membership in renderer.py's `_LOW_EXCLUDE_CLASSES`.

## Why

`_content_visible_at` already prefers a content class's `detail_visibility` attribute over the legacy exclude registries, so the outcome is **identical across all five detail levels** (visible at FULL/HIGH, dropped at LOW/MINIMAL/USER_ONLY) — the registry membership was simply a second, redundant source of truth for the same rule. This is the first built-in content type to use the class-attribute detail-visibility mechanism that plugins already rely on, keeping the rule next to the type it governs.

`ClassVar` is excluded from dataclass fields, so the constructor is unaffected. The pre-render `_filter_by_detail` whitelist (entry-level, before the factory exists) is unchanged; its comment now cross-references the class attribute.

## Verification

Full unit + snapshot suite **1916 passed, 7 skipped** (snapshots unchanged); `pyright` + `ruff` clean.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal message visibility control mechanism for improved alignment between visibility declarations and filtering operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/daaain/claude-code-log/pull/176?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/claude_code_log/models.py
+++ b/claude_code_log/models.py
@@ -2,7 +2,7 @@
 
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import Any, Union, Optional, Literal
+from typing import Any, ClassVar, Union, Optional, Literal
 
 from pydantic import BaseModel, Field
 
@@ -476,6 +476,12 @@ class AwaySummaryMessage(MessageContent):
     user comes back to a session. Distinct from HookSummaryMessage (tool noise)
     and the level-bearing SystemMessage (info/warning/error).
     """
+
+    # Recaps are narrative content, not noise: visible at FULL and HIGH,
+    # dropped at LOW and below (alongside bash/thinking). Declared via the
+    # class-attribute detail-visibility mechanism so the rule lives with the
+    # content type rather than in renderer.py's exclude registries.
+    detail_visibility: ClassVar[DetailLevel] = DetailLevel.HIGH
 
     text: str  # Recap prose; may contain light markdown.
 
--- a/claude_code_log/renderer.py
+++ b/claude_code_log/renderer.py
@@ -3169,14 +3169,13 @@ _HIGH_EXCLUDE_CLASSES: tuple[type[MessageContent], ...] = (
     UnknownMessage,
 )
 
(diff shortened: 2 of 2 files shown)


cboos:
> visible at FULL and HIGH, dropped at LOW and below (alongside bash/thinking)

Oh, that's an oversight from #141 (true, in #111, I wrote "at least the `--detail high level`, if not the `--detail low`, so it was ambiguous). 

But the intent was to **keep** them at a low detail level, because in itself it's high-level point of view summary of the activity that just happened. There's even a point in keeping them at user-only level, to really have a very high-level point of view of what happened: user said this, agent did that (just the recap, not the various messages).

Semi-OT: I keep getting confused with the "level"... I think we made the wrong call, semantically speaking. When you write `--detail high` and `--detail low`, it is clear enough. But I can't help to associate "high" with "seeing things from a high-level perspective" and "low" with "low-level details", that is, just the opposite. More on that in #159.

Back on point: I think we should always show the recaps, with a `--no-recaps` to eliminate them from `user-only` if that's not wanted there. That `--no-recaps` flag could also be useful for MINIMAL, if we don't want the "redundancy":

| You want to see: user | agent | recaps | How to achieve that? |
|--|--|--|--|
| ✅ | ✅ | ✅ | `--detail minimal` (and above if you also want more) |
| ✅ | ✅ |  | `--detail minimal --no-recaps` |
| ✅ |  | ✅ | `--detail user-only` |
| ✅ |  |  | `--detail user-only --no-recaps` |
|  |  | | `--detail session` (maybe?) |

cboos:
To conclude on this branch, which is only a refactoring... well, we'll keep the existing behavior, see #159 and #179 for follow-ups.

Now, onto the code, I wonder if instead of a special rules we couldn't simply have a `filter_by_detail` method on the Transcript classes?

Issue #129, opened 2026-04-25
Suppress noise in system info messages


cboos:
<img width="1159" height="348" alt="Image" src="https://github.com/user-attachments/assets/5bbd1692-6c55-45eb-8fd2-821dbedc6a44" />

Simplify to:

- ℹ️  `/status`
- ℹ️ Status dialog dismissed
- ℹ️  `/stats`
- ℹ️ Stats dialog dismissed

daaain:
I'm not even sure if the "dismissed" messages add any value, but I guess we can include them for completeness

cboos:
Well, I don't have enough practical experience yet, since the feature is very new, but the more I look at the `--detail high` output, the more I think it should probably become the default... The `--detail full` would still be there when we'd want to have a look at *everything* (in which case, all the hooks, the system information notices, etc. would have to be there).

cboos:
Another example:

<img width="1261" height="367" alt="Image" src="https://github.com/user-attachments/assets/ccb70b94-e6bd-4e43-b485-6fb52e8f096c" />

Bonus point if we make the system info notice "closer" to the slash command (maybe system info could always have the size of its parent?). I'm not sure if we could have a heuristic here for "pairing" the system-info with the slash command (pair-first, pair-middle, pair-last).

PR #184, opened 2026-05-30
Compute branch preview once from the DAG-line

cboos:
## Summary

The branch `SessionHeaderMessage` preview — which feeds the body header title, the session/graph nav, and the fork-point box's backlink, all of which must agree on `Branch • <uuid8> • <preview>` — was computed in two coordinating places:

1. The render loop extracted text from the branch's first transcript entry (the trigger that opened the branch).
2. An `_enrich_branch_titles` post-pass walked `ctx.messages` and back-filled the preview when (1) found nothing — the case where a branch starts with an assistant turn ("No response requested." after `/exit`) or a tool_result.

This collapses the two into one: `_extract_session_hierarchy` exposes each branch's full DAG-line uuid sequence (`hierarchy[sid]["uuids"]`); the render loop builds a `uuid → entry` map once; and `_build_branch_header` scans those uuids in DAG order for the first user entry with non-empty text via the same `extract_text_content` + `create_session_preview` path the slash-command capture uses. The post-pass goes away. `_branch_label` remains the single label formatter.

> **Stacking note:** this PR targets `wf/simplify/factor-session-headers` (the previous PR in the simplification track), not `main`. The base will be retargeted to `main` after the parent merges. Reviewing against the previous branch keeps the diff scoped to just this opportunity's change.

## #129 slash-command precedence — preserved structurally

The previous rule ("never widen an existing preview, including a 5-char `/exit` body, with a longer-but-less-informative scan result") only existed because `_enrich_branch_titles` could overwrite. With a single computation, DAG order does the work: a slash-command user entry sits at `uuids[0]`; the scan picks it on the first iteration and breaks. Length never enters the comparison.

## Implicit sidechain invariant

The deleted `_enrich_branch_titles` had an explicit `if msg.is_sidechain: continue` guard to avoid lifting a subagent's inner first user prompt as the branch preview (agent messages can carry a branch's `render_session_id`). The new code does not need that guard because it scans the branch's own DAG-line uuids; agent entries live in a separate `{trunk}#agent-{id}` DAG-line (re-parented by `_integrate_agent_entries` before `build_dag`). The guard was an artifact of the old `ctx.messages`-iteration approach.

## What changed

- `claude_code_log/renderer.py` — `_extract_session_hierarchy` adds `"uuids"` to each entry; `_render_messages` builds the `uuid → entry` map and threads it into `_build_branch_header`; `_build_branch_header` does the DAG-line scan; `_enrich_branch_titles` and its call deleted; nav-pass comment refreshed.
- `test/test_branch_label_source.py` (new) — two regression tests:
  - `test_branch_starting_with_assistant_uses_later_user_text` — the case the deleted enrich pass back-filled.
  - `test_branch_starting_with_slash_command_preserves_129_precedence` — slash-command body at branch root wins over later longer user reply.
- `dev-docs/dag.md`, `dev-docs/application_model.md`, `dev-docs/rendering-architecture.md` — replace `_enrich_branch_titles` references with the new `_build_branch_header` surface.

Net: `renderer.py` -83 lines; one new test module (+245); three doc files lightly touched.

## Test plan

- [x] `just test` — 1912 passed, 7 skipped (1910 prior + 2 new).
- [x] HTML + Markdown snapshot suites — byte-identical (zero `.ambr` churn — single-source picks the same preview as trigger + back-fill).
- [x] Fork / session / detail-level + async-agents suites green.
- [x] `test_branch_label_source` — 2 new tests pass via the real `load_directory_transcripts` → `generate_template_messages` pipeline.

Plan reference: `work/simplify-converter-renderer.md` §3 row 7, §4 "dev/simplify-branch-label-source".

## Summary by CodeRabbit

* **Refactor**
  * Branch header previews are now computed during header construction by scanning each branch’s message sequence for the first user text, removing a separate post-render backfill step.

* **Tests**
  * Added regression tests for preview selection edge cases: assistant-first branches, slash-command precedence, and ensuring agent/sidechain messages are not used for branch previews.

* **Documentation**
  * Updated architecture and model docs to reflect the new preview-generation ordering and message reindexing behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/daaain/claude-code-log/pull/184?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/claude_code_log/renderer.py
+++ b/claude_code_log/renderer.py
@@ -722,17 +722,6 @@ def generate_template_messages(
     with log_timing("Pair Skill tool_uses", t_start):
         _pair_skill_tool_uses(ctx)
 
-    # Branch headers were composed by ``_render_messages`` based on the
-    # branch's first message — but real branches sometimes start with an
-    # assistant entry (e.g. "No response requested." after a `/exit`),
-    # leaving the body header bare ``Branch • <uuid8>``. Walk the branch
-    # contents now that ``ctx.messages`` is final, lift the first
-    # UserTextMessage's text as the preview, and re-label the branch
-    # header. This keeps the body header, the session/graph index, and
-    # the fork-point box all carrying the same ``Branch • <uuid8> •
-    # <preview>`` string.
-    _enrich_branch_titles(ctx)
-
     # Populate junction forward links on fork-point messages
     with log_timing("Link junction forwards", t_start):
         _link_junction_forwards(ctx)
@@ -894,6 +883,13 @@ def _extract_session_hierarchy(
             "is_branch": dag_line.is_branch,
             "original_session_id": dag_line.original_session_id,
             "first_uuid": dag_line.uuids[0] if dag_line.uuids else None,
+            # Full DAG-line uuid sequence. ``_build_branch_header``
+            # scans this to find the first user entry with text for
+            # the branch preview — needed when the branch's first
+            # entry is an assistant turn (e.g. "No response requested."
+            # after ``/exit``) and the trigger-message-only path
+            # would leave the preview empty.
+            "uuids": list(dag_line.uuids),
         }
 
     junction_targets: dict[str, list[str]] = {}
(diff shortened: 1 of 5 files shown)


cboos:
@coderabbitai full review

cboos:
@coderabbitai full review

Issue #153, opened 2026-05-10
Minor CSS clean-ups


cboos:
(I got burnt by the "small fix doesn't deserve a PR" for 7c2e6f6, so to avoid that mistake again, here's a collection of "small fixes that don't deserve a PR on their own"...)

- [x] **Replace deprecated word-wrap with overflow-wrap in hook attachment blocks.** (from #149) 

The word-wrap property is deprecated. Replace with overflow-wrap: break-word; at line 407 to preserve behavior and clear the deprecation warning. Note: the same deprecated property appears at lines 470, 492, 746, 900, and 1006 elsewhere in the file—consider addressing all instances for consistency.

- [x] **too large spacing between "▼▼" in Chrome** (from #73)

Fix with `letter-spacing: -.2em` (not too much as then this would break FF)

- [x] **Fix WebFetch output** ("22 lines" / "559.5 KB" overlap)

<img width="1165" height="453" alt="Image" src="https://github.com/user-attachments/assets/5662033a-b0f8-4014-ad86-03b8ba00ed02" />

**more** ...

cboos:
The third unchecked item ("Fix WebFetch output — 22 lines / 559.5 KB overlap") is a specific instance of a more general shape that's recurring as plugins land. Surfaced again in PR #173's federation review by clmail/carol — her plugin emits a structured header `<p>` (From/To/Subject/Time) followed by `render_markdown_collapsible(...)`, and the fold-bar's first row visually overlaps the preceding paragraph because `.markdown p`'s bottom margin is zero and the `<details>` top margin is also zero. Even more severe than the WebFetch case since the overlapping content is structured (header lines) rather than a single URL/meta line.

The underlying class of bug: **`render_markdown_collapsible(...)` (and the bare `.collapsible-code` primitive generally) doesn't reserve any top margin when preceded by sibling content within the same `.markdown` container.** The existing fix for `.tool_result .collapsible-code { margin-top: -2.5em }` is a deliberate negative margin to merge the collapsible into the tool_result header bar; the WebFetch case overrode it back to `margin-top: 0` with `.webfetch-result.collapsible-code` (line 1233-1235) because WebFetch needs space for meta badges — but `0` is still wrong when there's actually a preceding paragraph (the bug in this issue's screenshot).

Suggest reframing the unchecked third item to cover the generalized shape:

[...]

Issue #90, opened 2026-02-06
Support async agents


cboos:
This is relatively new (seen in 2.1.34 today), and we kind of "support" it already, but it could be nicer.

<img width="1175" height="506" alt="Image" src="https://github.com/user-attachments/assets/1a3403f7-5cf9-422f-9151-880340c60eef" />

```
Async agent launched successfully.
agentId: a8b740b (internal ID - do not mention to user. Use to resume later if needed.)
The agent is working in the background. You will be notified automatically when it completes.
Continue with other tasks.
output_file: /tmp/claude-1000/-home-cboos-.../tasks/a8b740b.output
To check progress before completion (optional), use Read or Bash tail on the output file.
```
<img width="913" height="504" alt="Image" src="https://github.com/user-attachments/assets/9d0c2611-87bb-4f8b-9a7e-1784fd1297a4" />

```xml
<task-notification>
<task-id>a8b740b</task-id>
<status>completed</status>
<summary>Agent "Analyze relay.py coverage gaps" completed</summary>
<result>Now I'll analyze the uncovered lines and provide a comprehensive summary.

## Summary of Uncovered Lines in relay.py

### Coverage Analysis

The uncovered lines group into three main functions with varying levels of test coverage:
...
6. **Attachment handling** - Test file reading, base64 encoding, missing files, permission errors</result>
<usage>total_tokens: 23099
tool_uses: 2
duration_ms: 15506</usage>
</task-notification>
Full transcript available at: /tmp/claude-1000/-home-cboos-.../tasks/a8b740b.output
```

cboos:
Actually, the "Async agent" is just one async task among others... A similar mechanism is at play with async commands.

<img width="1252" height="845" alt="Image" src="https://github.com/user-attachments/assets/32bd88ae-3739-4c2e-84f6-5a62848ac211" />

Pairing: `Command running in background with ID: b0c03rdzj. Output is being written to: /tmp/claude-1000/-home-cboos-Workspace-.../d6a3e78a-8ce0-4afc-bbac-fc4b1ede655c/tasks/b0c03rdzj.output` with:

```xml
<task-notification>
<task-id>b0c03rdzj</task-id>
<tool-use-id>toolu_011CgEhqXBFYPepwFvrdbRXe</tool-use-id>
[...]

PR #187, opened 2026-05-30
Fix collapsible body overlapping preceding content in tool cards (#153)

cboos:
## Problem

In a WebFetch tool result the meta badge (`200` / `559.5 KB` / `1.5s`) renders **on top of** the collapsible's `▶ N lines` summary instead of stacked — the overlap shown in #153.

## Root cause

`.tool_result .collapsible-code { margin-top: -2.5em }` is a deliberate negative margin that tucks the *first* collapsible up under the tool's header bar. But it applies to **every** collapsible in the card, so any collapsible that follows sibling content gets dragged up over it:

| Case | Preceding sibling |
|------|-------------------|
| WebFetch | `.webfetch-meta` badge |
| Async Task answer (#90) | "Result (from async notification)" label |
| A header line emitted before `render_markdown_collapsible(...)` | a `<p>` header |

The existing WebFetch override `.webfetch-result.collapsible-code { margin-top: 0 }` was **dead code**: those two classes sit on different elements (the wrapper `<div>` vs. the nested `<details>`), so it never matched and `-2.5em` won.

## Fix

One general rule keyed on the structural discriminator — whether the collapsible's wrapper is the first child of the card body:

```css
.tool_result .content > * + * .collapsible-code { margin-top: 0.5em; }
```

`* + *` matches any card-body child with a preceding sibling; its collapsible then gets a small positive margin instead of the merge, while the first collapsible keeps the merge. A positive value (not `0`) also separates the collapsible from a zero-bottom-margin preceding paragraph, so it covers the header-before-collapsible shape generally, not just WebFetch. Removes the dead WebFetch override and the now-redundant `.task-async-answer` override (both subsumed; comments left pointing to the general rule).

## Before / after

- **Before:** meta badge + `▶ 33 lines` overlap.
- **After:** meta badge on its own line, `▶ 33 lines` cleanly below.
- A normal Read result is byte-for-byte identical before/after — its collapsible is the first child, so the header-bar merge is preserved.

## Tests

- `test/test_collapsible_overlap_browser.py` — Playwright geometry guards: asserts the WebFetch meta badge no longer overlaps the summary (fails on pre-fix CSS by ~32px), and that a first collapsible still tucks under the header bar.
- Snapshot HTML tests updated for the embedded-CSS text change; no structural HTML changed.
- `just ci` green (format, unit/TUI/browser tests, ruff, pyright, ty).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented collapsible panels from overlapping preceding content by applying consistent spacing for tool-result sections and removing special-case overrides.

* **Tests**
  * Added browser regression tests to validate collapsible layout and spacing: ensure meta badges no longer overlap collapsibles while preserving intended tuck behavior for first collapsibles.


cboos:
@coderabbitai full review

PR #188, opened 2026-05-30
Extract compute_session_data + compute_project_aggregates (C9b)

cboos:
## Summary

Consolidates session-metadata derivation that was previously inlined verbatim at four call sites into two pure helpers in `converter.py`:

- **`compute_session_data(messages, *, include_cwd=True, warmup_session_ids=None) -> dict[str, SessionCacheData]`** — shared prologue (summary/ai-title via the existing `prepare_session_*` helpers), warmup filter, agent-session coalescing via `get_parent_session_id`, per-session grouping, team_name capture, first-user-message preview, and requestId-deduped token derivation.
- **`compute_project_aggregates(messages) -> dict[str, Any]`** — project-wide token totals + earliest/latest bookend timestamps + `total_message_count`.

All four sites become thin call sites:

| Site | After |
|---|---|
| `_update_cache_with_session_data` | Wrapper: `compute_session_data` + empty-session filter + cache write; `compute_project_aggregates` + aggregate write |
| `_build_session_data_from_messages` | Three-line delegate to `compute_session_data(include_cwd=False, warmup=…)` |
| `_collect_project_sessions` | Pure projection adapter — projects `SessionCacheData` to the nav dict shape, collapses `summary`+`ai_title` via `prepare_session_*`, skips empty, preserves `timestamp_range` formatting + `[No user message found in session.]` placeholder |
| `process_projects_hierarchy` cache-unavailable fallback | The ~50-line inline aggregate loop becomes `compute_project_aggregates(messages)` + 6-line dict unpack |

> **Stacking note:** this PR targets `wf/simplify/session-scan-characterization` (the previous PR in the simplification track), not `main`. The base will be retargeted to `main` after the parent merges. Reviewing against the previous branch keeps the diff scoped to just this opportunity's change.

## Behavior changes — applied uniformly

Two locked behavioral changes are applied through the shared helpers so every call site agrees:

- **D1 — count un-keyed assistant usage.** The pre-refactor cache path used `request_id and request_id not in seen`, which silently zeroed assistant entries that had a real `usage` block but no `requestId`. The pagination fallback already counted them. Both paths + the index inline loop now apply the same rule: dedup *repeats* of a present `requestId`; count entries with no `requestId` always.
- **D2 — count `PassthroughTranscriptEntry` in `message_count`.** The pre-refactor cache path included Passthrough (no isinstance filter), the fallback excluded it (explicit isinstance filter). The shared helper no longer excludes Passthrough, so every site agrees: Passthrough participates in the DAG and contributes to `message_count`.

Both decisions are applied through the SINGLE shared helper that all four sites call — so uniformity is structural, not four parallel edits that could drift.

## Visible test flips (= the characterization safety-net firing)

The previous PR (#186) pinned the pre-refactor divergences. This PR flips the assertions to the new expected values, with the fixtures unchanged:

- fallback `s1.message_count`: 3 → **4** (D2 — Passthrough now counted)
- cache `s1.total_input_tokens`: 10 → **110** (D1 — un-keyed now counted), with the other three token totals flipping in lockstep
- index inline-loop project totals: 17/8/2/3 → **117/58/22/33** (D1)
- cross-equality on the D1/D2 fixture now HOLDS cache==fallback (was the gap this PR closes)

Two additional tests added per the previous PR's forward-looking note:

- `test_d1_d2_fixture_cache_equals_fallback` — pins the new invariant that the cache and fallback paths produce identical `SessionCacheData` on the D1/D2 fixture too (was previously the gap; typical-case equality was the only one held).
- `test_d1_d2_fixture_message_count_via_cache` — pins D2 at the index layer. The index-layer characterization test only pinned token totals; this drives `process_projects_hierarchy` with the cache mocked None and reads the cache the fallback's `_update_cache_with_session_data` populated, asserting the unified D2 rule (Passthrough counted) holds there too. Closes the silent-unpinned gap.

## Out of scope

- `SessionCacheData` model shape — unchanged.
- Cache schema, mtime / library-version invalidation — unchanged.
- `team_name` per-session capture — still the first non-None `teamName` seen; rides the shared helper now.
- Snapshots — byte-identical (zero `.ambr` churn). D1/D2 deltas affect token totals + message_count only, neither feeds rendered transcript HTML/Markdown.

## Test plan

- [x] `just test` — 1922 passed, 7 skipped (1920 prior + 2 new).
- [x] HTML + Markdown snapshot suites — byte-identical.
- [x] Targeted at-risk suites (pagination page-assignment, cache, dag_integration, ai_title, teammates_parsing) — 244 passed.

Net: `claude_code_log/converter.py` -220 lines; `test/test_session_scan_characterization.py` +97 lines.

## Summary by CodeRabbit

* **Refactor**
  * Centralized session/project metadata computation for consistent session summaries, message/token counts, and navigation display across cached and on-demand paths.
  * Ensures pagination fallback and project index totals match cached results and filters out sessions lacking a first user message.

* **Tests**
  * Expanded tests to assert cache vs fallback parity and to pin message/token counting behavior across processing paths.

### DIFF
--- a/claude_code_log/converter.py
+++ b/claude_code_log/converter.py
@@ -1048,101 +1048,127 @@ def _assign_sessions_to_pages(
     return pages
 
 
-def _build_session_data_from_messages(
+def compute_session_data(
     messages: List[TranscriptEntry],
+    *,
+    include_cwd: bool = True,
+    warmup_session_ids: Optional[set[str]] = None,
 ) -> Dict[str, SessionCacheData]:
-    """Build session data from messages when cache is unavailable.
+    """Compute per-session ``SessionCacheData`` from a flat message list.
 
-    This is a fallback for pagination when get_cached_project_data() returns None.
+    Shared by every site that needs per-session metadata — the cache
+    writer (``_update_cache_with_session_data``), the pagination
+    cache-miss fallback (``_build_session_data_from_messages``), and
+    the projects-index nav builder (``_collect_project_sessions``).
+    All three previously inlined slight variants of this loop; the
+    divergences were deliberately consolidated by opportunity 9 (see
+    ``work/simplify-converter-renderer.md``) onto two locked rules:
 
-    Args:
-        messages: All messages (deduplicated)
+    - **D1** (count un-keyed assistant usage): assistant entries with
+      a ``requestId`` are deduped via ``seen_request_ids`` (a retried
+      entry is counted exactly once); assistant entries *without* a
+      ``requestId`` are always counted. Both the cache and fallback
+      paths previously used distinct rules — the cache truthy-guard
+      (``request_id and request_id not in seen``) silently zeroed
+      un-keyed usage; the fallback already counted it. Unified here
+      to "count un-keyed everywhere".
 
-    Returns:
-        Dict mapping session_id to SessionCacheData
+    - **D2** (include ``PassthroughTranscriptEntry`` in
+      ``message_count``): Passthrough entries
+      (``type: progress``, ``agent-setting``, …) participate in the
+      DAG chain and contribute to project size. The cache path
+      already counted them today; the fallback excluded them. Unified
+      here to "count Passthrough everywhere".
+
+    Other invariants preserved:
+
+    - ``get_parent_session_id`` coalesces agent synthetic sessionIds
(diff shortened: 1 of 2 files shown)


cboos:
@coderabbitai full review

PR #190, opened 2026-05-31
Derive render_session_id from the SessionTree, not a loop variable

cboos:
## Summary

Replaces the loop-local `current_render_session` variable in `_render_messages` with a precomputed `uuid_to_render_sid` map built once from the `SessionTree` via the existing `session_hierarchy`. The map drives **both** `render_session_id` assignment **and** branch-header creation off the same trigger — the first surviving uuid in a branch DAG-line creates the header AND gets its `render_session_id` set, with no possibility of one happening without the other.

## Why

The loop variable carried a latent bug: `_filter_by_detail` can drop a branch's `first_uuid` at non-FULL detail (e.g., a `tool_use`-only assistant entry that opens a fork gets filtered at MINIMAL because all its content items are stripped). In that case, the branch-start trigger never fires; the loop variable stays at the previous branch's sid (or `None`) and every surviving branch entry that follows silently inherits a stale `render_session_id` — landing in the wrong session group in the final tree.

The map-driven approach reads grouping from the authoritative DAG, so the trigger fires off whichever branch entry survives, regardless of which one was originally first.

## Map shape

- **Branch lines** (`is_branch=True`, sid contains `@`) — every uuid → branch sid.
- **Agent lines** (sid contains `#agent-`) — every uuid → `parent_session_id` (the *immediate* parent; a nested agent B inside agent A maps to A's sid, not transitively to the trunk).
- **Trunk lines** — OMITTED. The caller's `map.get(uuid)` returns `None`, the `TemplateMessage`'s `_render_session_id` stays unset, and `TemplateMessage.render_session_id` falls back to `meta.session_id` at read time. Same behavior as the previous `not is_agent → current_render_session = None` reset.

## Removed

- `branch_start_uuids` dict (the `first_uuid → branch sid` lookup).
- `current_render_session` loop variable + its set on branch enter and reset on new trunk session.
- `agent_parent_session` inline derivation (now folded into the map's agent-line entry).
- The `effective_session = agent_parent_session or current_render_session` composition at every register site — replaced by a single map lookup at the top of the loop.

## Branch header trigger contract preserved

The main risk: "a branch message tagged to a branch sid with no header lands in the `_reorder_session_template_messages` unmatched tail." Fixed by unifying header creation and `render_session_id` assignment under the same map-driven check. The trigger now fires when:

```
(effective_render_sid is a branch sid)
  AND (branch not yet seen)
  AND (message's own sessionId is NOT an agent session)
```

That last clause keeps an agent-message-whose-parent-is-a-branch from incorrectly creating the branch header out of order. In practice the branch's own anchor entry (the Task tool_result that parents the agent) is itself a branch-line entry and always precedes the agent in the stream because `_integrate_agent_entries` splices agents at the anchor, so the branch header is always created by the anchor and the agent's later entries just look up the already-seen branch sid.

`_build_branch_header.first_uuid` field also updated: was `getattr(message, "uuid", "")` (trigger's uuid), now `b_hier.get("first_uuid")` (the canonical first uuid of the DAG line). The trigger may now be a later survivor rather than the original first; the canonical first uuid documents the branch on the header itself.

## Latent-bug regression test

`test/test_dag_integration.py::TestRenderSessionResetAcrossSessions::test_branch_first_uuid_filtered_out_still_groups_correctly`.

Constructs the exact shape the previous loop variable mishandled: trunk + a fork whose two assistant children have *only* a `tool_use` content item (no text). At MINIMAL, `_filter_by_detail` strips `ToolUseContent` and drops the entries entirely. The branch's surviving user message asserts its `render_session_id` is a branch sid (`s1@...`), not `s1` (trunk).

The test FAILS on `main` (the loop-variable bug — `c1.render_session_id == "s1"`) and PASSES with the map-driven derivation in place. Pins the structural fix.

The `TestRenderSessionResetAcrossSessions` class's existing `test_second_session_not_polluted_by_first_session_branch` test body is unchanged; the class docstring + sibling test are rewritten to reflect the new derivation.

## Test plan

- [x] `just test` — 1923 passed, 7 skipped (1922 prior + 1 new). No snapshot regenerations needed (existing fixtures don't include the filtered-first_uuid shape).
- [x] HTML + Markdown snapshot suites — 15 passed, byte-identical.
- [x] Fork/session/detail/async-agents suites — green.
- [x] Failing-on-main contract verified in a `/tmp` throwaway worktree.

Diff: `claude_code_log/renderer.py` (+116 / -59 net +57 — the map builder + its docstring + the unified-trigger comment outweigh the deletions). `test/test_dag_integration.py` (+177 / -4 — one new regression test + class-doc rewrite).

## Summary by CodeRabbit

* **Bug Fixes**
  * Messages now consistently render under their correct branch session IDs even when intermediate entries are filtered.
  * Branch headers anchor to the canonical branch start for stable rendering and filtering.
  * Trunk session headers are registered before branch headers when trunk messages are filtered, preserving header ordering.

* **Tests**
  * Added regression tests covering branch grouping, render-session assignment, and trunk-vs-branch header ordering under minimal/detail filtering.

### DIFF
--- a/claude_code_log/renderer.py
+++ b/claude_code_log/renderer.py
@@ -3886,6 +3886,45 @@ def _render_messages(
             and effective_session not in seen_sessions
             and not is_agent_session(msg_session_id)
         ):
+            # Ensure the branch's PARENT trunk session header is
+            # registered FIRST. At non-FULL detail every trunk
+            # message of the branch's parent session can be
+            # filtered out, leaving a branch descendant as the
+            # first surviving entry for that trunk session. Without
+            # this guard, the branch header would register before
+            # any trunk header → ``_reorder_session_template_messages``
+            # (which preserves session-header encounter order) would
+            # emit the branch header as a root section instead of
+            # nesting it under its parent trunk, and the branch
+            # header's ``parent_message_index`` would also be
+            # ``None`` because the trunk header isn't in
+            # ``ctx.session_first_message`` yet. Closes the
+            # CodeRabbit-flagged regression on PR #190.
+            #
+            # ``msg_session_id`` is the JSONL ``sessionId`` of the
+            # branch's own entry (the trunk session id, e.g. ``s1``
+            # — branch sids only exist in the in-memory DAG, never
+            # in the source JSONL). Use it as the parent trunk sid.
+            parent_trunk_sid = msg_session_id or "unknown"
+            if (
+                parent_trunk_sid
+                and parent_trunk_sid not in seen_sessions
+                and not is_agent_session(parent_trunk_sid)
+            ):
+                seen_sessions.add(parent_trunk_sid)
+                trunk_summary = sessions.get(parent_trunk_sid, {}).get("summary")
+                trunk_header_content = _build_trunk_header(
+                    parent_trunk_sid,
+                    trunk_summary,
+                    session_hierarchy,
+                    session_summaries,
+                    session_team_names,
+                    ctx,
+                )
+                trunk_header = TemplateMessage(trunk_header_content)
+                msg_index = ctx.register(trunk_header)
+                ctx.session_first_message[parent_trunk_sid] = msg_index
+
             seen_sessions.add(effective_session)
             branch_header_content = _build_branch_header(
                 effective_session,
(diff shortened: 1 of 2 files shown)


cboos:
@coderabbitai full review

cboos:
@coderabbitai full review

PR #193, opened 2026-06-01
Ghost-aware ctx.messages; skill-fold uses None slots (ghosting epic Phase 1)

cboos:
## Summary

First phase of the ghosting epic — a structural refactor that lets the renderer drop messages from the visible output **without** invalidating the index references that other passes rely on. See `work/ghosting-epic-plan.md` (included as part of this PR's commits) for the full design.

`RenderingContext.messages` is reshaped to allow `None` slots ("ghosts"). When a pass drops a message from the visible render path, it sets the slot to `None` instead of splicing it out of the list — every index around the dropped slot stays valid, so every stored index reference (`parent_message_index`, `session_first_message`, `junction_forward_links`, `pair_first/last/middle`) keeps working without a reindex.

This PR migrates the easier of the two `_reindex_filtered_context` callers (`_pair_skill_tool_uses`) to use the new mechanism. The detail-filter caller is Phase 2, after which `_reindex_filtered_context` can be deleted.

## Behavior

End-to-end output is byte-identical with main on every existing fixture (1924 passed, 7 skipped; HTML + Markdown snapshot suites 15 passed, zero `.ambr` churn). The migration is purely structural — same selection logic, same downstream rendering; the only difference is that consumed Skill-fold slots are `None`d in place instead of being spliced out + the message list reindexed.

> Caveat raised in review: existing snapshot fixtures don't exercise the Skill-fold path, so the byte-identity claim is narrow — see "Forward-looking note on snapshot coverage" below.

## What changed

- **`RenderingContext.messages: list[Optional[TemplateMessage]]`** — the slot type is the entire model change. No new field on `TemplateMessage`. Slots can be `None` to preserve indices around dropped messages.
- **`_visible(messages)`** — new iterator helper that yields non-`None` messages. Used by every reader of `ctx.messages` that shouldn't touch a ghost.
- **`_pair_skill_tool_uses`** — rewritten to set `ctx.messages[idx] = None` for each consumed slot instead of building a kept-list and calling `_reindex_filtered_context`. GC frees the dropped TemplateMessages immediately.
- **`_filter_template_by_detail`** and **`_reorder_session_template_messages`** widened to accept `Optional[...]` input and skip `None` slots; both return `list[TemplateMessage]` (the boundary points where ghosts are filtered out of the stream).
- **All ~16 direct `ctx.messages` iterators in `renderer.py`** switched to `_visible(...)` — trailing-link passes, task-metadata population, junction-forward linking, the in-loop scans inside `_build_branch_header`, etc.
- **`_reindex_filtered_context`** and its detail-filter call-site are **unchanged** for Phase 1 — Phase 2 deletes both.

## Plan divergence vs. `work/ghosting-epic-plan.md`

The original plan (§3.4–§3.6) proposed defensive ghost-skips in `_build_message_hierarchy`, `_build_message_tree`, and `_identify_message_pairs`. In implementation, `_reorder_session_template_messages` turned out to be the natural filtering boundary — its return type `list[TemplateMessage]` flows into all downstream passes, so they don't need ghost-awareness at all. **One filtering boundary, not seven defensive checks.** Pyright's narrow return type prevents Nones from ever reaching the downstream signatures, enforcing the invariant statically.

The plan's §3.4–§3.6 still document what *would* be needed if Nones flowed through every pass; the implementation uses the boundary approach. Recording the divergence here; plan update can fold in alongside Phase 2.

## New test (D12-prerequisite)

`test/test_skill_pairing.py::TestSkillFoldOnFork::test_skill_fold_inside_fork_at_full_keeps_branch_backref`

Exercises the PR #131 regression class through the ghosting path: a within-session fork whose Branch 1 contains a full Skill spawn (assistant Skill `tool_use` + user "Launching skill" `tool_result` + isMeta `sourceToolUseID` body). At FULL detail, asserts:

1. The Skill body is folded onto the `tool_use` (`skill_body` field set).
2. Exactly two `ctx.messages` slots are `None` — the consumed slash body + launching-skill `tool_result`. This is the new ghosting contract; pre-Phase-1 the consumed slots were deleted + reindexed (zero `None` slots), so the test fails on the pre-Phase-1 tip.
3. The Skill-bearing branch header resolves `parent_message_index` via `ctx.get(...)` to the actual fork anchor (uuid `c`), NOT to a phantom slot. This is the failure mode PR #131 introduced under the old reindex path; the verifier flagged it as missing coverage for D12, so landing it in Phase 1 means D12 (Phase 2 in the epic plan) inherits coverage when it lands.

Fixture quirk worth noting: two assistant siblings off the fork anchor (rather than the usual mixed user+assistant pattern) to bypass the DAG's `_stitch_tool_results` heuristic, which absorbs the common pattern as a tool-result side-branch. Inline comment in the fixture documents this.

## test_skill_pairing.py adaptation

The 16 existing tests in `TestSkillPairing` and adjacent classes iterate `ctx.messages` directly and assumed each element was a `TemplateMessage`. Patched their comprehensions / loops to filter `m is not None` first — same assertion intent, now ghost-aware. The pre-existing `TestReindexBranchBackrefs` class still exercises `_reindex_filtered_context` directly (Phase 2 will delete that test class along with the function it pins).

## Forward-looking notes from the monk review (non-blocking — flagged for Phase 2)

Both items raised during review; neither blocks Phase 1, both belong in the Phase 2 design.

### Anchor-target guard for Phase 2

Old `_reindex_filtered_context` *nulled* any reference pointing AT a consumed slot (`parent_message_index → None`, junction tuple dropped, `session_first_message` dropped). Ghosting instead leaves those references holding the ghosted index (a non-`None` int) and relies on `ctx.get()` returning `None`. Equivalent ONLY because, in Phase 1, every anchor index is derived via `_visible` or points at session-first messages — structurally never a skill-fold target. So nothing currently produces a ghosted anchor target.

The latent edge: `system_formatters.py:247-280` emits a raw `#msg-d-{parent_message_index}` anchor guarded only by `is not None` on the stored int — no `ctx.get()` existence check. If a later ghosting phase ever ghosts a slot that IS an anchor target, this silently becomes a dead anchor (where the old reindex would have suppressed the backlink). When Phase 2 deletes `_reindex_filtered_context`, either (a) assert `consumed_indices ∩ anchor-targets == ∅`, or (b) route the raw-anchor consumers through `ctx.get()` before emitting.

### Snapshot coverage gap for skill-fold paths

`message_id = f"d-{message_index}"` and `_build_message_hierarchy` does NOT renumber. Old code compacted `message_index` via reindex; ghosting keeps them sparse. A transcript that DOES contain a skill-fold now renders different (sparse, gapped) `d-{N}` anchor values + ancestry CSS classes than before this PR.

Functionally fine — `id` and `href` both use the same `message_index`, so internal nav stays self-consistent — but it means the 15 byte-identical snapshots confirm only that the *unchanged* paths are unchanged; none of them exercise a skill-fold, so the changed render path isn't snapshot-covered. The D12 test covers the data layer (None slots + backref via `ctx.get`) but not the rendered HTML. Cheap follow-up (Phase 2 or now): one snapshot fixture rendering a skill-fold transcript, to lock the anchor-consistency invariant (every `#msg-d-N` href has a matching `id`).

## Test plan

- [x] `just test` — 1924 passed, 7 skipped (1923 prior + 1 new D12-prerequisite test).
- [x] HTML + Markdown snapshot suites — 15 passed, byte-identical (zero `.ambr` churn). See caveat above re. skill-fold coverage gap.
- [x] Targeted: `test_skill_pairing.py` — 17 passed (16 patched + 1 new).

Net: `claude_code_log/renderer.py` +134/-58 (helper + model widening + skill-fold migration + reader audits); `test/test_skill_pairing.py` +219/-21 (None-aware patches + new D12-prerequisite test class).

The two planning-doc commits (`work/ghosting-epic-plan.md`) are included so reviewers see the full design alongside the implementation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve message positions with ghost placeholder slots so hidden entries no longer break indexes, anchors, previews, or backreferences.
  * Skill/tool folding now ghosts consumed messages in place and clears dangling anchor/parent references to avoid broken links or anchors.
  * Rendering and navigation now skip ghosted slots throughout preview, linking, and post-render flows; UI renders non-clickable labels for ghosted items.

* **Tests**
  * Added regression tests covering skill-folding, ghosting behavior, anchor repair, and forked-session navigation.

* **Documentation**
  * Added an epic rollout and test plan for the ghosting approach.

### DIFF
--- a/claude_code_log/html/templates/components/session_nav.html
+++ b/claude_code_log/html/templates/components/session_nav.html
@@ -15,9 +15,17 @@
         {% if session.is_fork_point is defined and session.is_fork_point %}
         <div class='session-nav-item session-fork-point'
             style='margin-left: {{ session.depth * 24 }}px'>
+            {% if session.message_index is not none %}
             <a href='#msg-d-{{ session.message_index }}' class='fork-link'>
                 &#x2442; {{ session.first_user_message }}
             </a>
+            {% else %}
+            {# Fork point was ghosted (e.g. a folded Skill slot) — show the
+               label without a dangling anchor. #}
+            <span class='fork-link'>
+                &#x2442; {{ session.first_user_message }}
+            </span>
+            {% endif %}
         </div>
         {% elif session.is_compaction_point is defined and session.is_compaction_point %}
         <div class='session-nav-item session-compaction-point'
--- a/claude_code_log/renderer.py
+++ b/claude_code_log/renderer.py
@@ -3419,6 +3419,41 @@ def _pair_skill_tool_uses(ctx: RenderingContext) -> None:
     for idx in consumed_indices:
         ctx.messages[idx] = None
 
+    # A branch header's ``parent_message_index`` and the
+    # ``session_first_message`` map are cached during ``_render_messages``
+    # — *before* this pass. If a fork point happened to land on a slot we
+    # just ghosted, that cached index now refers to a ``None`` slot and the
+    # rendered ``#msg-d-{N}`` backlink would dangle (the anchor href is
+    # emitted from the raw index; ``ctx.get()`` returning ``None`` doesn't
+    # suppress it). Null those stale refs so the anchor is simply omitted.
+    # ``junction_forward_links`` are populated *after* this pass
+    # (``_link_junction_forwards``), so they need no repair here.
+    _drop_anchor_refs_into_ghosts(ctx)
+
+
+def _drop_anchor_refs_into_ghosts(ctx: RenderingContext) -> None:
+    """Null anchor-target refs that now point at a ghosted (``None``) slot.
+
+    Keeps the index-stability contract honest for the only anchor sources
+    cached before ghosting runs: ``session_first_message`` and each branch
+    header's ``parent_message_index``. See ``_pair_skill_tool_uses`` for
+    why junction forward links are out of scope. Phase 2 of the ghosting
+    epic adds a broader ``_repair_stale_anchor_refs`` for the detail-filter
+    ghost path; this is the skill-fold-scoped counterpart.
+    """
+    ctx.session_first_message = {
+        sid: idx
+        for sid, idx in ctx.session_first_message.items()
+        if ctx.get(idx) is not None
+    }
+    for msg in _visible(ctx.messages):
+        if (
+            isinstance(msg.content, SessionHeaderMessage)
+            and msg.content.parent_message_index is not None
+            and ctx.get(msg.content.parent_message_index) is None
+        ):
+            msg.content.parent_message_index = None
+
 
 def _reindex_filtered_context(
     ctx: RenderingContext, filtered: list[TemplateMessage]
--- a/claude_code_log/renderer.py
+++ b/claude_code_log/renderer.py
@@ -1397,8 +1397,15 @@ def prepare_session_navigation(
                 insert_pos += 1
 
             # Fork point nav item — find the junction message and a
-            # meaningful preview (walk up past system hooks to find it)
-            fork_msg_idx = ctx.session_first_message.get(parent_sid)
+            # meaningful preview (walk up past system hooks to find it).
+            # Leave ``fork_msg_idx`` None until the fork point is actually
+            # located among the visible messages: if it was ghosted (e.g.
+            # a slash body or launch tool_result consumed by
+            # ``_pair_skill_tool_uses``), the ``_visible`` loop skips it and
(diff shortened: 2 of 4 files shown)


cboos:
@coderabbitai full review

cboos:
@coderabbitai full review

cboos:
@coderabbitai full review

cboos:
@coderabbitai full review

cboos:
@coderabbitai full review

PR #209, opened 2026-06-07
Keep recaps visible at all detail levels; add --no-recaps (#179)

### LINKED ISSUE #179: Recaps always visible (except when `--no-recaps`)
Initially an oversight from #141 which didn't follow the lead from #111 where I wrote "at least the `--detail high level`, if not the `--detail low` (but that was ambiguous). 

The idea is to **keep** them at all levels of detail, because a recap is in itself a summary  of the activity that just happened from a high-level point of view.

Therefore, I think there's even a point in keeping these recaps at the `user-only` level, to really have a very high-level view on what happened: user said this, agent did that (just the recap, not the various messages).

To get back the *really user-only* behavior, we can have a `--no-recaps` flag. That `--no-recaps` flag could also be useful for `--detail minimal`, if we don't want the "redundancy" between the agent messages and the recaps.

In summary:

| You want to see: *user* messages | *agent* messages | *recaps* | How to achieve that? |
|--|--|--|--|
| ✅ | ✅ | ✅ | `--detail minimal` (and above if you also want more) |
| ✅ | ✅ |  | `--detail minimal --no-recaps` |
| ✅ |  | ✅ | `--detail user-only` |
| ✅ |  |  | `--detail user-only --no-recaps` |
|  |  | | `--detail session` (maybe?) |

_Originally posted by @cboos in https://github.com/daaain/claude-code-log/issues/176#issuecomment-4577472473_

cboos:
Closes #179.

## Background

Recaps (`※ recap` / `away_summary` messages) are a high-level summary of the
activity that just happened. They were classified `detail_visibility = HIGH`,
so they were dropped at `--detail low` and below (an oversight from #141) —
exactly the levels where a one-line "user said this, agent did that" summary is
most useful.

## What this does

Keep recaps visible at **every** detail level, and add `--no-recaps` to
suppress them when you don't want them.

- `AwaySummaryMessage.detail_visibility`: `HIGH` → `USER_ONLY` (the
  least-verbose threshold), so recaps survive `full` → `user-only`.
- New `--no-recaps` flag, threaded through the renderers and converter. It
  suppresses recaps at any level, including `full` (so `--detail full
  --no-recaps` works). Honored by the HTML, Markdown, and JSON renderers.

Resulting matrix:

| user | agent | recaps | flags |
| :--: | :---: | :----: | :--- |
| ✅ | ✅ | ✅ | `--detail minimal` (and above) |
| ✅ | ✅ |    | `--detail minimal --no-recaps` |
| ✅ |    | ✅ | `--detail user-only` |
| ✅ |    |    | `--detail user-only --no-recaps` |

## Tests

- Recap visible at all five detail levels + `--no-recaps` suppresses at all five.
- End-to-end matrix test for the four rows above.
- Cross-format coverage: `--no-recaps` honored by html / markdown / json.
- Updated the minimal real-projects assertion to allow (and pin to a recap) the
  now-surviving recap.

`dev-docs/messages.md` and `application_model.md` updated.

## Deferred follow-up

The issue lists a `--detail session → (nothing)` row marked "(maybe?)". That
would be a new detail level below `user-only` that renders no messages — a
separate feature (new `DetailLevel` value + "render nothing" semantics),
deliberately **not** included here. Tracked as a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added a `--no-recaps` option to suppress recap/away-summary messages in exports.

* **Changes**
  * Recap messages are shown at all detail levels by default.
  * The `--no-recaps` behavior is applied consistently across HTML, Markdown, and JSON outputs and reflected in variant filenames.

* **Tests**
  * Expanded tests to cover recap visibility matrix and suppression across formats.

* **Documentation**
  * Guides updated to describe recap visibility and the new flag.

### DIFF
--- a/claude_code_log/cli.py
+++ b/claude_code_log/cli.py
@@ -706,6 +706,17 @@ def _validate_git_link_template(template: str) -> None:
     ),
 )
 @click.option(
+    "--no-recaps",
+    is_flag=True,
+    help=(
+        "Suppress '※ recap' (away-summary) messages. Recaps are otherwise "
+        "shown at every detail level — they are themselves a high-level "
+        "summary of activity (#179). Use this to get a 'really user-only' "
+        "view (--detail user-only --no-recaps) or to drop the recap/agent "
+        "redundancy at --detail minimal."
+    ),
+)
+@click.option(
     "--debug",
     is_flag=True,
     default=False,
@@ -735,6 +746,7 @@ def main(
     compact: bool,
     git_link: Optional[str],
     no_timestamps: bool,
+    no_recaps: bool,
     debug: bool,
 ) -> None:
     """Convert Claude transcript JSONL files to HTML or Markdown.
@@ -1004,6 +1016,7 @@ def main(
                 detail=detail_level,
                 compact=compact,
                 no_timestamps=no_timestamps,
+                no_recaps=no_recaps,
             )
             click.echo(f"Successfully exported session to {output_path}")
             if open_browser:
@@ -1064,6 +1077,7 @@ def main(
                 filter_path=filter_path,
                 write_combined=write_combined,
                 no_timestamps=no_timestamps,
+                no_recaps=no_recaps,
             )
 
             # Count processed projects
@@ -1123,6 +1137,7 @@ def main(
             update_cache=output is None,
             write_combined=write_combined,
             no_timestamps=no_timestamps,
+            no_recaps=no_recaps,
         )
         if input_path.is_file():
             click.echo(f"Successfully converted {input_path} to {output_path}")
--- a/claude_code_log/converter.py
+++ b/claude_code_log/converter.py
@@ -1286,6 +1286,7 @@ def _generate_paginated_html(
     session_tree: Optional[SessionTree] = None,
     detail: DetailLevel = DetailLevel.FULL,
     compact: bool = False,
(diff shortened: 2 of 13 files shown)


cboos:
(Claude) Good catch on the variant/cache identity — confirmed and fixed in a0ef294.

`--no-recaps` was indeed missing from `variant_suffix()`, so a `--no-recaps`
export collided with the plain one on both the output filename and the
cache/path-existence key — the second run would be served the stale prior
variant (same class as the #165 no-timestamps fix).

Rather than decorating the suffix string at each call site, I added `no_recaps`
as a parameter to `variant_suffix()` and threaded it through every caller
(combined / paginated / individual-session / single-session / hierarchy in
converter.py, plus the html/markdown/json renderers). One difference from the
suggested patch: because `--no-recaps` filters *messages* (not formatting like
compact/no-timestamps, which are Markdown-only), it earns a suffix slot for
**every** format — html and json included, not gated on `is_markdown`.
`VARIANT_ENTRY_RE` already matches the new `.no-recaps` segment.

Tests added: a `variant_suffix` matrix for `no_recaps` across html/md/json (+
composition with detail/compact/no-timestamps), and an end-to-end test that the
plain and `--no-recaps` exports land on distinct files. `just ci` green.

PR #225, opened 2026-06-20
feat: Add multi-provider session viewer support (Codex, Gemini, OpenCode, agy)

iansherr:
## Summary

This PR extends claude-code-log from a Claude Code-only viewer into a multi-provider session viewer that supports **5 AI coding tools**: Claude Code, Codex CLI, Gemini CLI, OpenCode, and Antigravity CLI (agy).

### Provider Abstraction Layer
- New `providers/base.py` — abstract `BaseProvider` with `discover_sessions()` and `load_session()` interfaces
- New `providers/registry.py` — auto-discovery and management of providers based on `~/.` directories
- 5 concrete adapters: `claude.py`, `codex.py`, `gemini.py`, `opencode.py`, `agy.py`

### CLI & TUI
- `--provider <name>` — filter to a specific provider
- `--list-providers` — show which providers are available
- `--all-providers` — generate unified cross-provider index
- TUI provider filter (`f` key cycles through providers)

### All-Providers Unified Index
- Generates a single `index.html` across all providers with session counts, provider badges
- Full-text search across all sessions (content previews loaded via `max_messages=10`)
- Clicking a search result navigates to the individual session HTML file with auto-search triggered via `#search=` URL hash
- Session cards link directly to individual HTML files

### Bug Fixes
- Fixed search next/prev buttons on index page (was trying to scroll to non-existent highlights)
- Fixed `navigateToMatch` to handle index page context correctly
- Fixed provider badge CSS for all-providers layout

### Provider Notes
- **agy**: Reads `brain/<uuid>/.system_generated/logs/transcript.jsonl` (not protobuf SQLite) — binary blobs in `conversations/*.db` are not human-readable
- **OpenCode**: Reads sharded JSON format (`session/*.json`, `message/{id}/*.json`)
- **Gemini**: Extracts from `logs/*.ndjson` with CLI command extraction
- All providers support `max_messages` for preview loading

### Default Output Directory
- Changed default output from `./claude-code-log-output` to `~/.local/agent_log_viewer/`
- Keeps generated files out of project directories, follows XDG convention
- `--output` flag still works as before

## Testing
- 2059 unit tests passing
- 8 snapshot tests updated (search template changes)
- Format/lint clean (ruff)
- Pyright: 274 pre-existing errors (no regressions)

## Scope Note

This is a substantial feature addition. I started with a fork of daaain/claude-code-log (MIT licensed) and extended it with the provider abstraction. The original Claude Code functionality is fully preserved — new providers are additive.

I'm happy to continue working on this (there are more polish items I'd like to address), but wanted to share the progress in case you'd like to review or have preferences on direction.

The output directory change is included here as a separate commit for easy extraction if you'd prefer it as a standalone PR.


daaain:
Wow, this is incredible, thanks a lot for taking the time implementing and submitting!

I opened an issue to discuss on a higher level, would appreciate your insights having already done this implementation: https://github.com/daaain/claude-code-log/issues/226

iansherr:
Thanks! Happy to see you're excited. Just responded in the issue discussion. Meantime, I'm going to keep tinkering with the fork, just because there are some other elements I've been itching for, like parallel processing from your todo list. I'll keep them as separate worktrees as much as possible so it's easy to pick and choose merges however you move forward.

cboos:
Hello Ian, this looks indeed promising. However, it seems your branch had some trouble with the symlink in the docs ([docs/development](https://github.com/daaain/claude-code-log/blob/main/docs/development)), bringing in a lot of unrelated (and outdated?) changes in. As a result, CodeRabbit went back to its burrow... Can you try to tidy this up a bit, so that only the relevant changes are included?

Also: `Pyright: 274 pre-existing errors` - this is also concerning... I wonder why, as main is "clean" in that respect.

cboos:
(Claude) Thanks for this — it's an ambitious contribution and the multi-provider direction is compelling. The core abstraction is well-placed: `BaseProvider` adapting each format into the existing `TranscriptEntry` model means the whole downstream pipeline (DAG, factories, rendering) is reused unchanged. A few things would need addressing before this can land, plus a fidelity question.

### Type checking — fails the `pyright` CI gate
`uv run pyright` reports **274 errors, all in the new/changed code** (≈223 across the provider adapters, 40 in the new `generate_all_providers_index` functions, ~10 in the CLI block); the pre-existing code is clean. They're mostly loose typing — untyped JSON access (`entry.get(...)` into bare `dict`/`list`), `tool_calls: list` without an element type, and one `reportPrivateUsage` (`registry._provider_classes` accessed outside the class). `just ci` runs pyright as a gate, so this currently makes CI red. The tests do pass, but the type gate doesn't.

### Docs symlinks got replaced with copies
`docs/development`, `docs/changelog.md`, and `docs/contributing.md` are **symlinks** in the repo (to `dev-docs/`, `CHANGELOG.md`, `CONTRIBUTING.md`). This PR replaces them with ~5000 lines of copied files, which will silently drift from their sources. This looks like a checkout artifact (e.g. a Windows clone materializing symlinks) rather than an intentional change — worth restoring the symlinks.

### Shared index template
The all-providers search support modifies the shared index-page JavaScript rather than isolating it, so the standard (Claude-only) index now carries the extra code too. Behavior is preserved (the new branches are inert without provider data), but it's worth being aware the change isn't fully isolated.

### Real-data test of the `agy` adapter
Tested against a real Antigravity CLI session. It renders and reads coherently — good. Two fidelity gaps surfaced:

1. **Flat DAG.** Every entry is created with `parentUuid=None`, so the DAG builder logs `N roots found (N-1 unexpected), walking all from earliest` and renders a flat list — none of the structured rendering (tool cards, threading, nesting) engages. Threading entries sequentially (each entry's `parentUuid` = the previous one's `uuid`) would let the existing renderer do much more.

2. **Dropped entry types.** `_parse_entry` only handles `USER_INPUT`, `PLANNER_RESPONSE`, `CHECKPOINT`, and `LIST_DIRECTORY`. `GENERIC`, `RUN_COMMAND`, `VIEW_FILE`, and `CODE_ACTION` fall through and are silently skipped — in one real ~47-entry session that dropped **19 entries (~40%)**, including all command output, file views, and code edits. You see the planner's narration and tool-call *invocations*, but not the tool *results*. Either mapping these to `tool_use`/`tool_result` pairs or at least emitting them as text would preserve the content; a warning on unrecognized types would also help.

### Suggestion
The abstraction is the right shape, but with 5 adapters at once (each a flat, lossy MVP) this is a lot to land and verify together. It might be smoother to land the abstraction plus **one** fully-realized provider — properly typed, threading entries, and modeling tool calls/results — as the quality bar, then add the others incrementally against that template.

cboos:
(Claude) One more, found while testing the CLI end-to-end: **the `--provider <name>` option is a no-op.** It's declared (a `click.Choice` of `claude/codex/gemini/opencode/agy/all`) and documented ("Use --provider to limit to a specific provider"), but `main()` never branches on the `provider` argument — only `--list-providers` and `--all-providers` are actually wired. So `--provider agy` (or any single-provider value) is silently ignored and falls through to the default Claude path. Either wire it to a single-provider discover+render, or drop the option so it isn't an advertised flag that does nothing.

(Minor, and likely not worth chasing given the data is old: on a real multi-provider system the `codex` and `gemini` adapters report as available — `is_available()` only checks that the top-level dir exists — but discover zero sessions, silently, because their assumed on-disk layout doesn't match the real one, e.g. codex walks `sessions/<year>/<month>/<day>/rollout-*.jsonl` while an actual session can sit flat in `sessions/`. A discovery that finds nothing under an "available" provider could at least warn.)

cboos:
(Claude) Related gap: `--all-providers` / `--provider` should compose with the path-output flags (`--expand-paths` / `--filter-path`, i.e. "Obsidian mode") the way `--all-projects` does — currently they don't. `generate_all_providers_index` doesn't accept those params, so they're silently ignored under `--all-providers`; and passing `--all-providers --all-projects` together doesn't help because the `all_providers` branch returns before the `all_projects` one. (Worth deciding what path-filtering means for non-Claude providers, whose sessions aren't organized by project path.)

cboos:
Superseded by #242.

Issue #192, opened 2026-06-01
Parse memory?


daaain:
I can't check now, but wondering if we have a message in the log for Claude interacting with [auto-memory](https://code.claude.com/docs/en/memory#auto-memory).

If so, do we also need to parse the files in the memory directory?

cboos:
Yes, but these are regular Read/Write/Edits... 

<img width="1295" height="1050" alt="Image" src="https://github.com/user-attachments/assets/2b9c9d8f-26b5-464d-a029-3d00d003c43e" />

Not sure if there's much more to do. I suppose we could detect if the targeted file corresponds to a memory file and then "flag" the tool somehow (extra 🧠 icon?)

daaain:
Ah, right, so it looks like the memory "system" is basically just a few prompts (and a lot of these results are using the hosted agents memory and dreaming): https://github.com/search?q=repo%3APiebald-AI%2Fclaude-code-system-prompts%20memory&type=code

I guess that would explain why it's so random, I rarely spotted Claude using it, except when it retrieved an outdated memory and started following an old pattern that I needed to talk it out of 😅

Rendering a bit differently would be nice, but I'm mainly interested in being able to filter these to analyse usage, maybe we could match on the project directory path and memory both being present in these tool paths to prevent false positives?

cboos:
We could use `"Write/Edit/Read memory <short-path>"` for tool_use title, when the path matches the `$CLAUDE_PROJECT_DIR/memory/<short-path>` pattern.

cboos:
@daaain: a first shot can be seen in #204; but I won't have time to iterate on it until Sunday (same thing for my other PRs, they're going to stay WIP until I come back...).

cboos:
<img width="1234" height="830" alt="Image" src="https://github.com/user-attachments/assets/0e999b6b-c17d-4103-b498-7cc26ef3806f" />

(the link was missing the `memory/` part, hence today's churn)

PR #234, opened 2026-06-23
Render Read/Write of any Markdown file as Markdown, not Pygments

### LINKED ISSUE #232: Specialize Write tool for Markdown files
As a follow-up to the MEMORY topic (#192), and in general, for any fully contained Markdown file (hence, Write), I think we could just render the Markdown the usual way, instead of pygmentizing it.

The question is more for Read: for full reads, we could probably do that as well, but for partial reads, we could possibly land within a code block and get garbled output... 

We already do that for MEMORY files, but why not for all .md files?

cboos:
Closes #232.

## Problem

A follow-up to the auto-memory work (#192). We already render auto-memory file bodies as rendered Markdown instead of Pygments-highlighted source — but that specialization was keyed on the *memory path*, not on the file being Markdown. Any other `.md` file (a README, a doc, a plan) read or written during a session still rendered as highlighted source.

## Change

Generalize the body rendering to **every** Markdown file:

- New `is_markdown_path()` (`.md`/`.markdown`, case- and separator-insensitive) in `html/utils.py`. Memory paths are a subset: `is_memory_path ⊂ is_markdown_path`.
- **Write** of any `.md` → always rendered as Markdown. A Write carries the file's full content, so there's no partial-slice concern.
- **Read** of a `.md` → rendered as Markdown only for a **full** read (`start_line == 1 and not truncated`, via `_is_full_read`). A partial slice (`offset`/`limit`) can begin or end mid-code-fence and would render garbled, so partial reads keep Pygments — where a line-numbered source view is more useful for a slice anyway.
- Reuses the HTML-escaping markdown helper (`render_user_markdown_collapsible`): file content is untrusted regardless of being a memory file, so raw `<script>`/HTML renders as text, not live DOM.

### Deliberately unchanged

- **Memory** keeps its extra specialization: the 🧠 title, relative-link resolution, and always-Markdown body (even for partial reads). General `.md` files get neither the 🧠 title nor link rewriting.
- The **Markdown output renderer** is untouched (still fenced code blocks) — embedding raw Markdown-in-Markdown risks heading bleed and fence collisions, and the issue is framed around HTML pygmentizing.
- **Edit** stays as-is (diff view + partial snippet).

## Testing

New `test/test_markdown_file_rendering.py` (15 tests): `is_markdown_path` (`.md`/`.markdown`, case, Windows, memory-subset, negatives incl. `.mdx`); Read full→Markdown / partial→Pygments / truncated→Pygments / non-md→Pygments / escaping / no-link-resolution; Write md→Markdown / non-md→Pygments / escaping / no-link-resolution. The full/partial/truncated split is pinned behaviorally through the public `format_read_output` (no private-symbol import). `test_memory_rendering.py` regressions all pass — memory behavior unchanged. No snapshot churn (no existing fixture does a non-memory `.md` Read/Write).

`just ci` green (unit + TUI + browser + snapshots, ruff, pyright, ty).

## Summary by CodeRabbit

* **New Features**
  * Improved Markdown rendering: complete `.md`/`.markdown` reads and writes now display as formatted Markdown with safer HTML escaping.
  * Enhanced memory file handling, including Markdown link resolution.
  * Partial or truncated Markdown reads continue using syntax-highlighted rendering to avoid corrupted Markdown structure.

* **Tests**
  * Added coverage for Markdown path detection (including Windows paths and case-insensitivity), read/write rendering behavior, and link rewriting rules.

### DIFF
--- a/claude_code_log/html/tool_formatters.py
+++ b/claude_code_log/html/tool_formatters.py
@@ -23,6 +23,7 @@ from typing import Any, Optional, cast
 
 from .utils import (
     escape_html,
+    is_markdown_path,
     is_memory_path,
     render_collapsible_code,
     render_async_result_body,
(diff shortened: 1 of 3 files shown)


cboos:
@coderabbitai full review

Issue #221, opened 2026-06-18
`--output` to an existing file silently skips regeneration and keeps stale content


samestep:
### Summary

If the `--output` path already exists and its embedded version marker matches the current `claude-code-log` version, regeneration is **skipped** — even when the source is a *completely different* transcript. Worse, the CLI still prints `Successfully converted <new source> to <path>` while the file retains the **old** content. This is a silent data-correctness bug, not just a no-op.

This is why my personal notes say "make sure there's no file already at the path you're writing to, since otherwise it seems to just do nothing" — I now have to `rm -f log.md` before every run.

### Repro (v1.4.0)

```
# Convert A -> out.md
claude-code-log A.jsonl -f markdown -o out.md
md5sum out.md   # -> 25987e8a...

# Convert a DIFFERENT file B into the SAME path
claude-code-log B.jsonl -f markdown -o out.md
# prints: "MARKDOWN file out.md is current, skipping regeneration"
# prints: "Successfully converted B.jsonl to out.md"
md5sum out.md   # -> 25987e8a...  (STILL the content from A.jsonl!)
```

### Cause

The single-file / combined path (`convert_jsonl_to`, `converter.py:1718-1762`) gates the write on `should_regenerate`, which calls `renderer.is_outdated(output_path)`. `is_outdated` (`markdown/renderer.py:2169`, and the HTML/JSON equivalents) only compares the `<!-- Generated by claude-code-log vX -->` version marker — it has no notion of *which source* produced the file. So a same-version file at the user's chosen path is treated as "current" regardless of content.

Note the `--session-id` path (`generate_single_session_file`, `converter.py:2239`) *always* overwrites, so behavior is also inconsistent between the two ways of producing a single file.

### Suggested fix

When the user passes an explicit `--output`, always (re)generate — the staleness/version-marker heuristic is only meaningful for the tool's own managed `combined_transcripts*`/`session-*` artifacts, not for an arbitrary user-chosen destination. The CLI already special-cases `-o` for cache writes (`update_cache=output is None`, `cli.py:1123`); the regeneration skip should be bypassed the same way.

---
*Investigation and write-up by Claude (Claude Code).*

PR #253, opened 2026-06-30
Regenerate single-file output when its source grows; fix success-message accuracy (#221 follow-up)

cboos:
## Summary

Follow-up to #221/#237. Two user-visible staleness/messaging bugs in the
default single-file conversion path, plus message-accuracy fixes surfaced
while reviewing the change.

### Bug 1 — stale HTML on a grown single-file source
`claude-code-log <file>.jsonl [--detail low]` writes `<file>[.low].html`, then
re-runs were skipped purely on the embedded **tool version**
(`renderer.is_outdated`), which ignores source freshness. So an in-progress
session that grew between runs was wrongly reported "current" and served stale
HTML. #237 only forced regeneration for the explicit `-o file` case.

**Fix:** add a source-mtime term to the single-file/no-cache branch —
regenerate when the source file is newer than the existing output. Scoped to a
file source (a directory source has no single mtime and is handled by the
cache path). The version-marker check is kept for tool-version bumps.

### Bug 2 — misleading success message
On a skip, the CLI printed `Successfully converted …` on top of the
converter's `… is current, skipping regeneration`. The converter now reports
whether it actually (re)wrote output via a non-breaking optional `regenerated`
out-param (the `Path` return is unchanged — ~20 callers unaffected); the CLI
prints the success line only on a real regeneration. Applied symmetrically to
the directory branch (`Successfully combined …`).

### Message-accuracy fix
The regeneration gate initially tracked only the **combined** write, so
`--combined no` (which writes only per-session files, never a combined) fell
completely silent despite producing every session file — and the per-session
count was dropped whenever the combined was current but a session was
rewritten. The gate now also counts `_generate_individual_session_files`'s
return, i.e. "combined **or** any session file (re)written". Pure skips still
stay quiet.

## Tests
- `TestSingleFileStaleness` (test_output_explicit.py): grown source regenerates;
  unchanged source skips with no redundant success line; directory no-op skip
  suppresses the combined success line; `--combined no` confirms its
  per-session work.
- Flipped `test_single_file_mode_regeneration_behavior`, which previously
  enshrined the stale behavior as "expected".

All four behavioral tests were verified RED on the prior code and GREEN after.

## Not addressed here
`--no-cache` on a **directory** with a grown source can still serve stale HTML
(the source-mtime term is scoped to file sources). This is pre-existing, not
introduced here, and tracked separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated CLI conversion messaging to show “Successfully …” only when regeneration actually occurs, with distinct output for single-file, directory runs, and `--combined no`.
  * Improved regeneration detection for combined/paginated outputs and refined directory reporting when only session files are regenerated.

* **Tests**
  * Strengthened single-file staleness/skip assertions to regenerate only when the source JSONL is newer and to avoid extra success lines when skipped.
  * Added coverage for directory/session-only regeneration scenarios, including missing sessions and `--combined no` behavior.

### DIFF
--- a/claude_code_log/cli.py
+++ b/claude_code_log/cli.py
@@ -1346,15 +1346,25 @@ def main(
             click.echo(f"Successfully converted {input_path} to {output_path}")
         else:
             jsonl_count = len(list(input_path.glob("*.jsonl")))
+            session_suffix = ""
             if write_individual:
                 ext = get_file_extension(output_format)
                 session_files = list(input_path.glob(f"session-*.{ext}"))
+                session_suffix = (
+                    f" and generated {len(session_files)} individual session files"
+                )
+            if write_combined:
                 click.echo(
-                    f"Successfully combined {jsonl_count} transcript files from {input_path} to {output_path} and generated {len(session_files)} individual session files"
+                    f"Successfully combined {jsonl_count} transcript files "
+                    f"from {input_path} to {output_path}{session_suffix}"
                 )
             else:
+                # `--combined no`: only per-session files were written. Don't
+                # claim to have "combined" anything, nor name a combined file
+                # that wasn't produced — report the per-session work instead.
                 click.echo(
-                    f"Successfully combined {jsonl_count} transcript files from {input_path} to {output_path}"
+                    f"Successfully processed {jsonl_count} transcript files "
+                    f"from {input_path}{session_suffix}"
                 )
 
         if open_browser:
--- a/claude_code_log/cli.py
+++ b/claude_code_log/cli.py
@@ -1332,12 +1332,17 @@ def main(
             force_regenerate=output is not None and _output_path_is_file(output),
             regenerated=regenerated,
         )
-        # On a skip the converter already printed "is current, skipping
-        # regeneration"; suppress the redundant/misleading success line.
-        was_regenerated = regenerated != [False]
-        if input_path.is_file():
-            if was_regenerated:
-                click.echo(f"Successfully converted {input_path} to {output_path}")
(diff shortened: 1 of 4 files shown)


cboos:
(Claude) The branch was rebased onto current `main` (4df1997) and two review nits were folded in since your last review (head is now `2fd72ef`):
- `--combined no` success line is now write_combined-aware ("Successfully processed …" instead of a false "Successfully combined …").
- A regeneration test's `time.sleep(0.1)` was replaced with `os.utime` strict mtime ordering (coarse-FS safety — your earlier nit).

@coderabbitai full review

cboos:
(Claude) Addressed the review of `2fd72ef` in `845cbb3`:

- **Coarse merged signal → false "combined" (Major):** replaced the single `list[bool]` with a `RegenerationReport(combined_regenerated: bool, sessions_regenerated: int)`. The CLI now gates "Successfully combined … to `<file>`" strictly on the combined write, and reports "Successfully processed …" when only session files were rewritten — so it no longer contradicts the "combined … is current, skipping regeneration" line.
- **Paginated `did_regenerate=True` unconditional (Major):** `_generate_paginated_html` now returns `(first_page_path, wrote_any)`, so the combined flag is accurate when all pages are current.
- **Nitpick:** the per-session count now reflects files actually regenerated this run, not a glob of all `session-*.html`; added `test_combined_current_but_session_regenerated_says_processed` (verified RED before the fix).

`just ci` green; all `--combined` modes + the mixed case verified by running the CLI.

@coderabbitai full review

Issue #174, opened 2026-05-29
Add support for dynamic workflows


cboos:
See [Orchestrate subagents at scale with dynamic workflows - Claude Code Docs](https://code.claude.com/docs/en/workflows).

We should have most of the pieces in place (sub-agents, teammates, etc.), so I suppose it'll be about the orchestrator script... We'll see once we have a transcript example to analyze.

cboos:
Example:

> The heart of the system (claude_code_log/converter.py and claude_code_log/renderer.py) is still quite complex and somewhat difficult to follow, despite my earlier simplification and refactoring attempts (well, mostly in renderer.py so far, as you can see in what we documented in dev-docs/). It would be great to conduct a thorough review of the code, understand the algorithms we have in place deeply, the multiple passes, and find ways to refactor this to *simplifly* what we do (in particular, the session vs. rendering, where there are many cross dependencies at different levels, which I failed to simplify so far). To this end, you can prepare a dynamic workflow, roughly starting with isolating the simplification opportunities in various key places in the code, of course without breaking the tests (they can be adapted if needed, but keep their spirit) and likewise, without breaking the spirit of what we support (the DAG/tree of messages, the useful hierarchy structure, all what we documented in dev-docs/). You can create worktrees and coordinate using many branches that you merge progressively, ... but I don't want to be too directive about the *how*, I trust you to know how to proceed to keep the code changes understandable, well justified, and progressive!

(churn)

**Tool use:**

```json
{
  "parentUuid": "95dacc8e-deaf-4209-9d95-c55e5659f4a1",
  "isSidechain": false,
  "message": {
    "model": "claude-opus-4-8",
    "id": "msg_01LsFURfxVqKJGujo5e7pwqN",
    "type": "message",
    "role": "assistant",
    "content": [
      {
        "type": "tool_use",
        "id": "toolu_01C6kRkVbUro2HFY4axZtGJ9",
        "name": "Workflow",
        "input": {
          "script": "export const meta = {\n  name: 'simplify-converter-renderer',\n  description: 'Map converter.py/renderer.py, propose simplifications, adversarially verify, synthesize a sequenced plan',\n  phases: [\n    { title: 'Map', detail: 'parallel deep readers map each area + propose simplifications' },\n[...]"
        },
        "caller": {
          "type": "direct"
        }
      }
    ],
    "stop_reason": "tool_use",
    "[...]",
    "diagnostics": null
  },
[...]

PR #258, opened 2026-07-02
Handle the full variety of Workflow invocation shapes (#174 follow-up)

cboos:
## Summary

The #174 dynamic-workflow rendering assumed the inline-`script` invocation shape. Real sessions use others — a reference session's 6 `Workflow` tool_uses were **all `scriptPath`** invocations (one with `args`), which produced bare headers (no source, no description, invisible args) and a spurious `meta format may have drifted` warning per run per field. One run there also failed before launching any agent, leaving a snapshot but no journal — journal-led discovery skipped it entirely, so its `failed` status and error appeared nowhere.

- **Parse** (`workflow.py`): surface the terminal snapshot's full enrichment on `WorkflowRun` — `script` (the source that actually ran: the recovery route for non-inline shapes), `scriptPath`, `args`, `summary`, `error`, `defaultModel`, `durationMs`, `totalToolCalls`. The header resolves from the *effective* script (input, else snapshot copy), falls back to the snapshot `summary` for a missing description, and warns only when a **non-empty** script fails a parse the snapshot can answer — absent source is not drift.
- **Meta string parse**: accept any JS quote style (`'`/`"`/backtick) with backslash-escape support — real meta descriptions contain `\'`, which truncated the old parse mid-string.
- **Model** (`models.py`): `WorkflowToolInput` declares the invocation union — `script`, `scriptPath`, `name`, `args`, `resumeFromRunId` — instead of dropping the non-inline fields into `extra="allow"`.
- **Render** (HTML + Markdown): script body falls back to the snapshot copy; a `workflow-invocation` line shows saved-workflow name / scriptPath reference / resumed run id; `args` renders via the hybrid params table (HTML) / fenced JSON (Markdown).
- **Failure surfacing**: discovery also yields snapshot-only runs (no run dir/journal); a non-`completed` status renders as a chip next to the workflow name and the snapshot `error` as a collapsed fold — in both output formats.

## Verification

- New fixture `test/test_data/workflow_scriptpath/` (checked-in generator, same convention as `workflow_basic`): a `scriptPath`+`args` run whose snapshot carries the script (meta description with an escaped quote) + a snapshot-only `failed` run.
- 24 new tests in `test/test_workflow_rendering.py`; `just ci` green on every commit; snapshot `.ambr` churn is the new CSS only.
- End-to-end against the real reference session: all 6 invocations now render their orchestrator + invocation line, the failed run shows its status chip + stack trace, and the drift warnings are gone.
- `dev-docs/workflows.md` updated in-PR.

## Summary by CodeRabbit

* **New Features**
  * Workflow runs now show richer details, including invocation info, optional arguments, and clearer error output.
  * Support was added for multiple workflow invocation styles, including saved workflows and script-path based runs.

* **Bug Fixes**
  * Improved handling of failed workflow runs, including cases where only a snapshot is available.
  * Workflow header details are now more accurate when scripts contain quotes or special characters.

* **Documentation**
  * Updated workflow docs to reflect the expanded run lifecycle and new display behavior.


cboos:
If I may add a little note of self-satisfaction, our dynamic workflow rendering really nails it.
I rendered a relatively complex session with several dynamic workflows, and the HTML report clearly exposed all the details of what happened during those workflows.

@daaain, maybe it's time to cut a new release?

Issue #248, opened 2026-06-29
Runtime of tests is 3x to 6x slower on Windows than it is on Linux


cboos:
This started to become annoying... I got Claude to think about a few approaches.

cboos:
(Claude) Write-up of the investigation and the PRs it produced.

## TL;DR — three PRs, split by concern so they can land independently

- **#249 — test fixtures (test-only):** shrink the oversized fixture
  transcripts in the *copied* data; skip symlink tests where unavailable.
  `test_integration_realistic.py` **−38%**, full unit suite **−21%**
  (local Windows).
- **#250 — pytest-xdist per-worker import cost (test config):** stop
  importing `playwright`/`textual` and the pytest-playwright plugin on
  every unit worker. Collect-only **−26%**; per-worker savings that
  compound under Windows' process model.
- **#251 — SQLite cache build (production code):** reuse one DB
  connection per build instead of ~189 opens; `synchronous=NORMAL`. Cache
  build **−25%**, connection opens **189 → 14**.

## How the time actually splits (the surprising part)

A profiling pass on the worst offenders (the `test_integration_realistic.py`
family — 18 of the 25 slowest tests) showed:

1. **It is not the fixture copy.** `shutil.copytree` of the ~20 MB /
   ~100-file fixture set measured ~0.15s; teardown ~0.10s. The natural
   hypothesis — that the function-scoped `copytree` was the cost — was
   **refuted**. The entire per-test cost is `process_projects_hierarchy`
   (render + cache build), re-run over the large transcripts every test.

2. **Two distinct causes drive the Windows/Linux _ratio_** (as opposed to
   absolute slowness):
   - **pytest-xdist worker startup.** Windows *spawns* workers and
     re-imports the whole module graph per worker; Linux *forks*. A single
     trivial test costs ~4.5s at `-n0` vs ~18.6s at `-n auto` (12 workers)
     — ~14s of pure spawn + re-import. Module import alone is ~12.9s.
     Browser/TUI test modules, plus the auto-loaded pytest-playwright
     plugin, were imported on every unit worker even though unit tests
     never use them.
   - **SQLite connection churn.** The cache opened ~189 connections per
     single-project build (every method, reads included), each ~3.9ms —
     heaviest on Windows, where connection/file overhead is highest.

## Ideas tried — what worked, what didn't

| Idea | Verdict |
|---|---|
| Share the copytree across tests (function→module scope) | **Dropped** — copytree was never the cost (~0.15s). |
[...]

PR #264, opened 2026-07-04
Collect coverage only on the primary CI job; use the sys.monitoring coverage core

cboos:
Part of #248.

## Problem

All 10 CI matrix jobs ran pytest with `--cov` (xml + html + term, across the unit/TUI/browser step chain), but only the primary job (ubuntu 3.12) uploads the reports and feeds the PR coverage comment — the other 9 computed coverage and discarded it. Measured cost of those flags on the CI unit-test selection: **+67% pytest wall time** under the default C tracer (worst on Windows, where the unit step ran 734s vs 161s on ubuntu, run 28703179425).

## Changes

- **ci.yml**: coverage flags move to job-level `COV`/`COV_APPEND` env vars that resolve to empty on non-primary jobs, preserving the `--cov-append` chain (unit → TUI → browser → benchmark) on the primary job. Workflow validated with actionlint.
- **pyproject.toml**: `core = "sysmon"` in `[tool.coverage.run]` — the sys.monitoring core (Py 3.12+) cuts the remaining coverage tax on the primary job from +67% to +14%, with byte-identical coverage totals (84.71%). On Py < 3.12 coverage warns once and falls back to the default core; moot in CI since the only coverage job is on 3.12. Side benefit: de-flakes `test_render_performance_under_threshold`, which can exceed its threshold under the C tracer. `COVERAGE_CORE=ctrace` still overrides for A/B timing.
- **Coverage floor** bumped to 7.9 (introduced the `[run] core` setting; older versions warn and silently keep the slow core). uv.lock already resolves 7.12.0.

## Does dropping Windows coverage lose anything?

No. Diffing covered lines (full local Windows run vs the primary job's coverage artifact at the same commit) shows Windows uniquely covers exactly **5 lines**, all in `cli.py`: the SIGUSR1 no-op return and the drive-letter branch of `convert_project_path_to_claude_dir`. Those were already misses in the published report, which has always been ubuntu-only — so reported coverage changes by exactly 0 lines. (If cross-OS merged coverage is ever wanted, the groundwork is noted in the ci.yml comment; `relative_files = true` is already set.)

## Expected effect

- Non-primary jobs (all 5 Windows ones included): ~65% pytest wall-time cut on the unit step, plus removal of the ~40–50s coverage-combine/report tail after each of the three pytest steps.
- Primary job: coverage tax drops from +67% to +14%.

## Verification on this PR's CI run

- [ ] A non-primary job's pytest invocation shows no `--cov` flags in the log
- [ ] The primary job still uploads both coverage artifacts and posts the PR coverage comment
- [ ] Windows unit step wall time drops from ~10–12 min toward ~4–5 min

## Summary by CodeRabbit

* **Chores**
  * Updated the CI workflow to centralize test coverage settings and apply them only on the primary test job.
  * Adjusted coverage tooling settings to use a newer coverage engine and aligned the minimum supported coverage package version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,18 @@ jobs:
             python-version: "3.12"
             is-primary: true
 
+    env:
+      # Collect coverage only on the primary job: it is the only one that
+      # uploads the reports and feeds the PR comment, while --cov costs
+      # ~65% extra pytest wall time (worst on Windows). Windows-only
+      # covered lines measured at 5 (diffing covered lines per file,
+      # Windows vs the primary artifact): the SIGUSR1 no-op return and the
+      # drive-letter branch of convert_project_path_to_claude_dir, both in
+      # cli.py. Those were never in the published report (non-primary
+      # coverage was computed and discarded), so this changes nothing.
+      COV: ${{ matrix.is-primary && '--cov=claude_code_log --cov-report=xml --cov-report=html --cov-report=term' || '' }}
+      COV_APPEND: ${{ matrix.is-primary && '--cov=claude_code_log --cov-append --cov-report=xml --cov-report=html --cov-report=term' || '' }}
+
     steps:
     - uses: actions/checkout@v4
       with:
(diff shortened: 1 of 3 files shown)


cboos:
Windows 3.12 and its 21m... Fable called that a "lemon VM" and losing at the Windows runner lottery :P

But hey, the Linux numbers did drop again, so I guess we're good with these changes (initially aimed at reducing the gap between Windows and Linux build times!)

cboos:
(Claude) Re the zizmor `template-injection` nit on the `${{ env.COV }}` interpolations: declining, deliberately.

- **No injection surface**: `COV`/`COV_APPEND` are workflow-literal ternaries over `matrix.is-primary` — a boolean defined by the matrix itself. No attacker-controllable context (`github.event.*`, `inputs`, etc.) can reach these values, which is the threat the zizmor rule targets.
- **The suggested `$COV` form has a cross-shell trap in this matrix**: default `run:` shell is bash on ubuntu but **pwsh on windows-latest**, where `$COV` is an undefined PowerShell variable (env vars are `$env:COV`). It would only work today by coincidence (all Windows jobs happen to want the empty value); if a Windows job ever runs coverage (e.g. a future cross-OS coverage merge), the flags would silently vanish. `${{ env.* }}` expands before either shell parses the script, so it behaves identically across the whole matrix — that shell-agnosticism is why it was chosen.

Making shell expansion safe would need `shell: bash` on all four steps — more CI change than the trivial-severity finding warrants.

PR #265, opened 2026-07-04
Move TMP/TEMP to the workspace disk on Windows CI runners

cboos:
Part of #248; follow-up to #264.

## Problem

Even with coverage scoped to the primary job (#264), Windows unit-test steps still ran ~3.5x slower than ubuntu (~555s mean vs ~161s for the same selection). An A/B experiment on real windows-latest runners (run [28715118956](https://github.com/daaain/claude-code-log/actions/runs/28715118956), 5 variants x 2 reps, full evidence in `work/win-io.md`) isolated the cause: **TMP/TEMP default to the throttled C: OS disk**, and this tmp_path-heavy suite pays its I/O prices.

| variant | mean (s) | verdict |
|---|---|---|
| baseline | 554.5 | control |
| defender-off | 557.7 | no-op — real-time protection is already off on current images |
| defender-excl | 672.7 | no-op + VM variance |
| ReFS dev drive | 296.4 | helps, but dominated |
| **TMP/TEMP → RUNNER_TEMP** | **150.2** | **ubuntu parity** |

## Change

One early, Windows-gated step writing `TMP`/`TEMP` = `RUNNER_TEMP` (the workspace disk) to `GITHUB_ENV`, so every later step benefits (unit, TUI, browser, and dependency installs). Plus `work/win-io.md` recording the experiment so the step isn't "simplified away" later.

Side benefit: tmp dirs and the checkout now live on the same drive, removing the cross-drive layout that can break `os.path.relpath` between tmp_path and repo paths on Windows.

## Expected effect

Windows unit step ~735s (pre-#264) → ~150s; Windows jobs should drop off the CI critical path. This PR's own CI run is the end-to-end verification.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows CI reliability and performance by routing temporary files to a faster temporary location before install and test steps run.
* **Documentation**
  * Added a write-up summarizing the Windows CI performance experiment, findings, and the recommended configuration change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,21 @@ jobs:
         # only and needs the commits in the object DB).
         fetch-depth: 0
 
+    - name: Move TEMP to the workspace disk (Windows)
+      # On windows runners TMP/TEMP default to the C: OS disk, whose
+      # throttled I/O dominates this tmp-heavy suite; RUNNER_TEMP lives on
+      # the same disk as the workspace (D:). Measured via A/B experiment
+      # (run 28715118956): this redirect alone took the unit selection from
+      # 555s mean to 150s — parity with ubuntu (~161s). Defender is NOT the
+      # culprit: real-time protection is already off on current images
+      # (exclusions/disable measured as no-ops), and a ReFS dev drive
+      # (~296s) was dominated by this redirect. Details: work/win-io.md
+      if: runner.os == 'Windows'
+      shell: pwsh
+      run: |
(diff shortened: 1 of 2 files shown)


daaain:
Oh wow, this was unexpected!

cboos:
> Oh wow, this was unexpected!

Well, that was Fable continuing where Opus dropped the gauntlet ;-)

cboos:
For the record, the builds for this commit on main:

image

We're still slower on Windows (4-5 min) vs. Linux (2-3 min), but both are indeed in the same league now.

PR #268, opened 2026-07-07
Fix perpetual re-regeneration of projects with up-to-date caches

daaain:
Running the full generation twice in a row kept "regenerating" the same projects every run even though nothing changed. Reproduced on a real 2GB archive copy: 7 projects re-rendered their combined pages and 2 projects re-flagged one session on every single run, adding ~70s per run. Two root causes, both self-inconsistent staleness bookkeeping:

1. Paginated combined HTML (reason: message_count_changed). _generate_paginated_html cached len(page_messages) — the raw rendered entry count, which includes Summary/AiTitle/Attachment entries — but is_page_stale() compares that against SUM(sessions.message_count), computed by compute_session_data, which skips those types. Any project where the two counting rules diverge was permanently stale. Fix: cache the sessions-table sum for the page (same source the check reads), exactly as last_timestamp already did. The rendered count still feeds the on-page stats display.

2. Ghost sessions (reason: not_cached). A fork session whose own messages all deduplicate into the original session survives only through its agent sidechains ({trunk}#agent-{id} sessionIds). compute_session_data coalesces those to the trunk and writes a sessions-table row, but _generate_individual_session_files dropped agent ids from its render set instead of coalescing — so the session was never rendered, its html_cache row never written, and get_stale_sessions flagged it forever. Fix: coalesce with get_parent_session_id in the render set (generate_session already accepts {sid}#agent-* entries), and likewise in the pagination session filter so such sessions claim a page slot instead of their messages silently vanishing from the combined output.

Both fixes self-heal: the first run after upgrading regenerates the affected projects once, writes consistent bookkeeping, and every subsequent run is fully cached. Verified on the archive copy: run 1 regenerated the 9 known offenders, run 2 reported all 63 projects cached (8.4s total, previously ~80s with 9 spurious regenerations).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed pagination cache accounting so stored message totals match staleness checks, preventing unnecessary regeneration loops.
  * Improved per-page cached session tracking to keep message counts consistent across runs.
  * Updated session grouping to normalize agent-generated session IDs back to trunk sessions and exclude warmup-only sessions for both paginated and individual outputs.
  * Corrected team-name-per-session grouping to use the same consistent session normalization rules.

### DIFF
--- a/claude_code_log/converter.py
+++ b/claude_code_log/converter.py
@@ -1689,10 +1689,19 @@ def _generate_paginated_html(
         total_output_tokens = 0
         total_cache_creation_tokens = 0
         total_cache_read_tokens = 0
+        # The cached count must be the sessions-table sum, because
+        # is_page_stale() compares it against SUM(sessions.message_count)
+        # on the next run. len(page_messages) counts what's rendered
+        # (including Summary/AiTitle/Attachment entries that
+        # compute_session_data skips), so caching it left any page where
+        # the two rules diverge permanently "message_count_changed" —
+        # regenerating on every single run.
+        page_sessions_message_count = 0
 
         for session_id in page_session_ids:
             if session_id in session_data:
                 s = session_data[session_id]
+                page_sessions_message_count += s.message_count
                 if s.first_timestamp and (
                     first_timestamp is None or s.first_timestamp < first_timestamp
                 ):
@@ -1778,7 +1787,7 @@ def _generate_paginated_html(
             html_path=html_path,
             page_size_config=page_size,
             session_ids=page_session_ids,
-            message_count=page_message_count,
+            message_count=page_sessions_message_count,
             first_timestamp=first_timestamp,
             last_timestamp=last_timestamp,
             total_input_tokens=total_input_tokens,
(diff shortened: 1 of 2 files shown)


cboos:
Great! Strangely, I had no idea this could be a bug... even though I had >10 minutes to go for each repeated full generation (I thought that was just my `.claude/projects` being a bit large).

I tried again, 3 successive runs:

- Processed 797 projects in 370.7s;  31 projects updated
- Processed 797 projects in 206.0s;  5 projects updated
- Processed 797 projects in 198.7s; 1 projects updated

So, it's definitely much better, but maybe it would be worth an extra pass at analysis on my end.

daaain:
I wasn't sure what it was so just gave it to Fable as a goal to prevent it happening, but it's basically anything that messes with the reconciliation of the cache and files because we're filtering some stuff. I now have a sandboxed development env that I'm going to open source soon and pasted my entire ~/.claude/projects directory in there for it to play with and test.

Together with the other PR, cached stuff should now take a few seconds to reconcile unless there are some more bugs like this!

cboos:
Well, these 3 runs and the final ~200s were done with the latest main (e76e1a6), so it's definitely not just a few seconds for me...

daaain:
Got it, wasn't trying to gaslight you 😅 but with my transcripts it's fast now, so I was just explaining my not very sophisticated process which will need your projects directory

cboos:
> Well, these 3 runs and the final ~200s were done with the latest main ([e76e1a6](https://github.com/daaain/claude-code-log/commit/e76e1a697c49d3b671129e764f64d8ddceacea21)), so it's definitely not just a few seconds for me...

See #271 (doh!) and the more subtle #270.

daaain:
Nice! What's great is that once we get it down to seconds, it'll be obvious in the future if something breaks

PR #269, opened 2026-07-07
Filter <local-command-caveat>-wrapped caveat messages (Claude Code ≥ ~2.1)

daaain:
Newer Claude Code versions wrap the local-command caveat in a <local-command-caveat> XML tag instead of emitting the bare "Caveat: ..." text, so is_system_message() stopped matching it. The wrapped caveat then leaked into session previews / page titles (showing as the "first message" of any session that opens with a slash command) and rendered as a slash-command message in transcripts. It also promoted command-only sessions (e.g. a lone /model invocation) into the session index, which previously had no eligible first user message and were skipped.

Match the tag prefix in is_system_message and bump the cache compatibility floor so caches with the caveat baked into first_user_message rebuild on upgrade.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message filtering so caveat messages wrapped in special tags are now correctly treated as system messages, excluding them from conversation previews and generated output while keeping regular quoted text intact.
* **Documentation**
  * Added clearer notes about cache compatibility behavior for older versions.
* **Tests**
  * Added coverage to verify wrapped caveat filtering and ensure surrounding normal messages are still rendered.


claude_code_log/cache.py
```diff
@@ -1086,6 +1086,11 @@ def _is_cache_version_compatible(self, cache_version: str) -> bool:
             # JSONLs are re-ingested, so the project index keeps showing
             # the old session-id title until the cache is rebuilt.
             "1.2.0": "1.3.0",
+            # 1.5.0 filters <local-command-caveat>-wrapped caveat messages
+            # (Claude Code ≥ ~2.1) from session previews: caches built
+            # earlier have the caveat text baked into first_user_message
+            # for affected sessions.
+            "1.4.0": "1.5.0",
```
cboos, line 1093: Glad to see 1.5.0 is brewing ;-)

daaain: I realised we should have done it ages ago 😅 should probably cut a release once a week after every few PRs!

daaain: @cboos also wanted to email you about publishing to PyPI, but the address on your Github profile bounced my email. Mine is working, please drop me a message!

Issue #233, opened 2026-06-23
Fix confusing fork corner case with user messages children


cboos:
I saw this:

<img width="1327" height="1163" alt="Image" src="https://github.com/user-attachments/assets/61b156b2-b67d-41d7-b645-2d70a03f772d" />

Fork point `57e7a875-25a`

```
4481:{"parentUuid":"170e001c-9aa8-4d00-af5e-a8f95025e251","isSidechain":false,"type":"system","subtype":"turn_duration","durationMs":67152,"messageCount":1695,"timestamp":"2026-06-22T17:02:08.905Z","uuid":"57e7a875-25a3-4e84-884a-f73b3bfb73bd","isMeta":false,"userType":"external","entrypoint":"cli","cwd":"/home/cboos/SAM","sessionId":"700af82d-e4a8-4e25-81bd-46d62cb6e5d6","version":"2.1.185","gitBranch":"master","slug":"harmonic-plotting-reef"}
```

First child:

```
4482:{"parentUuid":"57e7a875-25a3-4e84-884a-f73b3bfb73bd","isSidechain":false,"promptId":"4c413cc1-5d66-4eda-a9b3-ffe4f69d4717","type":"user","message":{"role":"user","content":"I\nA practical concern here: ..."},"uuid":"5c12f81b-810e-48d8-9385-8a24b6c7a442","timestamp":"2026-06-23T08:35:36.855Z","permissionMode":"auto","origin":{"kind":"human"},"promptSource":"typed","userType":"external","entrypoint":"cli","cwd":"/home/cboos/SAM","sessionId":"700af82d-e4a8-4e25-81bd-46d62cb6e5d6","version":"2.1.185","gitBranch":"master","slug":"harmonic-plotting-reef"}
```

Second child:

```
4487:{"parentUuid":"57e7a875-25a3-4e84-884a-f73b3bfb73bd","isSidechain":false,"promptId":"37039ede-6fc4-4190-9e3c-f6e90e5f1c38","type":"user","message":{"role":"user","content":"I'll reply on all these points separately later on..."},"uuid":"6beaf80e-a77c-4f3b-bec2-733866ecc806","timestamp":"2026-06-23T08:39:15.173Z","permissionMode":"auto","origin":{"kind":"human"},"promptSource":"typed",...}
```
The thing is, the fork point "system/turn_duration" doesn't appear in the transcript (like goes to `#msg-d-2` which is the session header message).

[...]

PR #273, opened 2026-07-12
Rework in-page search: / shortcut, search-as-filter, highlight + unfold

### LINKED ISSUE #143: Improve search
The custom search is a bit wonky sometimes, I'm not sure what happened, but might have got broken at some point 🤷 

I'm not sure if I like that it hijacks the browser search, I'd probably change it to not use the same keyboard shortcut.

In terms of actual function, it could filter messages that contain the search term and highlight? Also, expand them all.

We should also add some browser tests to make sure it doesn't regress.

daaain:
Fixes #143

- Bind search to `/` (GitHub-style) and stop hijacking the browser's native Ctrl/Cmd+F find; Esc clears, Enter/Shift+Enter and F3/Shift+F3 navigate matches.
- Search now acts as a filter on transcript pages: non-matching messages get `search-hidden` (strict by default), matches get `search-match` with highlights, and their folded ancestors are auto-unfolded via a new `window.claudeLogRevealMessage` hook exposed by the fold machinery. A "Show context" toggle keeps ancestors of matches visible but dimmed (`search-context`).
- Fix regex-mode highlighting: a global pattern's lastIndex advanced by .test() leaked into matchAll(), silently skipping highlights.
- Fix findSessionForMessage for the nested DOM (session headers are ancestors now, not preceding siblings).
- Guard the filter MutationObserver against feedback loops from search's own class writes (compare class sets minus search-*).
- Move the search options (Regex / Visible only / Show context) out of the results-info block so they're usable before the first match.
- Mirror search filtering in the timeline: per-item hiding via timeline-filtered-hidden, applied live and on timeline init.
- Remove unused legacy templates search_inline_script.html and search_results_panel.html.
- Add browser test suite test/test_search_browser.py (11 tests) and document the runtime state classes in dev-docs/css-classes.md.
- Add work/archive-search-server.md: design draft for full-archive search from the SQLite cache via a stdlib-only local server.

## Polish

- Hint the `/` shortcut on the floating 🔍 button tooltip and show
  Shift+Enter / Enter/F3 key labels on the prev/next match buttons.
- F3 with no active query opens search like `/` (with a query it keeps
  navigating matches).
- Open collapsed `<details>` (tool results, structured params — nested
  included) on the path to every highlight, so a match inside a folded
  block is actually visible instead of silently matching.
- navigateToMatch targets the first *visible* highlight — the first one
  in DOM order can be the .preview-content duplicate that hides once
  its details is auto-opened.
- Emit data-uuid (stable transcript UUID, unlike positional msg-d-N
  slot ids) on message cards, and use it to keep the current match
  pinned when the same query re-runs after an option/filter change —
  toggling "Show context" no longer jumps back to match 1. This also
  settles the "anchor problem" in work/archive-search-server.md
  (option 1, decided & done).
- 3 new browser tests; dev-docs/css-classes.md updated.

## Summary by CodeRabbit

- **New Features**
  - Added a persistent “Show context” toggle for transcript search.
  - Improved index/search experience, including clearer “/” shortcut hints and revised navigation controls.

- **Bug Fixes**
  - Invalid regex inputs now reliably clear search results and filters.
  - Search now preserves the currently selected match across option changes and better auto-scrolls/wraps match navigation.
  - Transcript matches unfold as needed, and the timeline view stays synchronized with search filtering.

- **Documentation**
  - Documented the runtime search/filter visual states and stable message anchoring.

- **Tests**
  - Added browser automation coverage for search, context mode, unfolding, and timeline synchronization.


cboos:
> Bind search to / (GitHub-style) and stop hijacking the browser's native Ctrl/Cmd+F find

:tada: I'll test this evening!

daaain:
@cboos I'm particularly pleased with the feature that using the injected message uuid as `data-uuid` we know which message you were at when stepping through the results so if you change the filters it'll scroll where you were and add / remove messages around! Would have been a bit overkill for this, but we need the stable ids later for the db backed global search, so brought it forward

cboos:
Review from me:

- I like it!
- When you search with next/prev, you get to the next message, but there can be several hits per message; not a big deal, as the highlight seems to work reliably;
- The same prev/next (shift-)Enter key bindings only work when the focus is in the search box; OK as well, once you figure it out, but you need to figure it out first!

Review from Opus:

1. Low/hardening: the index results panel writes href="${escapeHtml(match.link)}", but escapeHtml doesn't escape " — an attribute-context gap (narrow reachability: needs a project dirname decoding to a quote; and the PR actually improved this line vs base, which had no escaping). Cheap fix: assign via the a.href DOM property.
2. Low/known limitation: no guard against catastrophic-backtracking user regexes — can freeze the tab on a huge page. Self-inflicted and local; a caveat or length cap would be belt-and-suspenders.
3. Info: performSearch sets isSearching=true without try/finally — an unexpected throw would silently kill search until reload.

=> finding 1 (with its one-line fix) and 3 are the ones worth a comment; 2 is fine as a "known limitation" remark.

cboos:
> @cboos I'm particularly pleased with the feature that using the injected message uuid as `data-uuid`

Well, I'm not that happy... Sure, using uuid is justified for the reasons you mention. But we already had `data-message-id`, also `id` (same value) and now `data-uuid`! I wonder if it wouldn't be possible to get rid of one, or more. What if we'd use id=uuid (+ suffix -n for unicity) only?
And that would provide stable URIs as well.

*core diagnosis: data-message-id is the real duplicate (same positional value as id, one single JS consumer — the fold machinery's getMessageEl — trivially migrated to getElementById).*

I'm arguing the case of a data-uuid/id merge as well, as the model stated *the uuid's semantic literally cannot be an id without a lossy positional -n suffix* - but pointing out that the split is deterministic. Better now:

---
You're right, and I'll concede the point with the verification to back it: `chunk_message_content` (renderer.py:1704) is a **pure function of the entry's own content list** — `[text, image, thinking, text, text, tool_use]` always chunks to the same 4 cards, and JSONL entries are immutable once written. So a suffix defined as the **chunk index within the entry** (not a render-occurrence counter, which is what monk and I were implicitly attacking) is content-derived and deterministic. It doesn't shift when the session grows, when history is rewound and replayed (new sequence, same uuids), or — nicely — across `--detail` variants, since chunking happens before the detail filter drops cards, whereas today's `msg-d-N` renumbers in *all* of those cases.

So the corrected trade-off table for `id='m-{uuid}-{k}'`:

**Now in favor:** stability across growth, rewind, *and* variants; the archive-search "anchor problem" solved natively (no resolver indirection needed); one identifier scheme for real messages.

**Still real, but bounded:**
1. **Migration cost** — the ~10 anchor-generating sites across 5 modules currently key on stored integer indices; they'd need the target's (uuid, k), i.e. the indirection you've already priced in. The fork-slot repair logic keyed on `msg-d-N` stability migrates too.
2. **Synthetic cards** need a parallel scheme, but it's actually *tidier* than I first said: session headers have a natural stable key (`session-{sid}`), and the #233 ghost fork-point placeholders carry the dropped node's uuid. So the fallback isn't a wart, it's arguably more semantic.
3. **Old `#msg-d-N` links break** — unavoidable, one-time; a transition release could dual-emit both ids if that matters.
4. **One assumption to verify empirically before committing**: that no uuid ever renders as cards in two places (fork/branch display duplicating an entry would break (uuid,k) uniqueness). The dedup/DAG work strongly suggests one card set per uuid, but this is exactly the kind of claim that wants an execution check over forky fixtures rather than confidence.

My updated position: this is a defensible design, not just defensible — the stability win is the *same* property Daniel is chasing in `work/archive-search-server.md`, so it might genuinely interest him as the endgame, with "drop `data-message-id`" as the uncontested first step either way. If you want, I'll have monk run the uniqueness probe (render the fork-heavy fixtures, assert no duplicate (uuid, chunk-kind-sequence) card sets) so the proposal you'd float upstream rests on verified ground rather than our shared belief in the dedup invariants.

cboos:
Quick experiment:

---
Commit 1 fe15d5b — drop data-message-id (a pure duplicate of id; migrated getMessageEl→getElementById). Small, green, independently landable. This is the redundancy that actually bothered cboos.

Uniqueness gate — PASSED: threaded chunk_index, checked (uuid,k) across all fork/dedup/sidechain fixtures + 101 real-project files (4357 uuid cards) → zero duplicates, zero ghost collisions.

Commit 2 f57deeb — card id → m-{uuid}-{k} via a post-render resolution pass, data-uuid dropped (uuid now in the id). All green (2339 unit / 77 browser / 8 snapshot, pyright+ruff clean), anchor parity verified.

Honest verdict to cboos:
1. Land C1 now — clean win, zero risk.
2. Gate C2 on archive-search-server being real — the stable-anchor benefit has no current consumer, and C2's small source diff (~93 lines) hides the real cost: the generators lack ctx, so a production impl needs per-generator resolver plumbing (~8 formatters/4 modules), not my post-render regex shortcut; plus 9 test files recoupled.

---

I can do a PR for C1 or add it here already (to compensate for the introduction of data-uuid).
And for C2, we can indeed revisit after archive-search-server.

daaain:
This sounds great @cboos, definitely in favour of merging the uuid into the id as we'd get stable anchors! Feel free to add to this branch!

The next/prev stepping messages was intentional because stepping between hits is what the browser native search does, but I guess we could do another keyboard shortcut to step inside the messages as maybe there's value in only having the filtered messages and then stepping through those (to differentiate from browser native)?

cboos:
> This sounds great @cboos, definitely in favour of merging the uuid into the id as we'd get stable anchors! Feel free to add to this branch!

Added.

> The next/prev stepping messages was intentional because stepping between hits is what the browser native search does, but I guess we could do another keyboard shortcut to step inside the messages as maybe there's value in only having the filtered messages and then stepping through those (to differentiate from browser native)?

Well, I think it's understandable to have "our" next/prev to find the *message*. The within message search is then "visual" (as the highlight works again) or via the restored CTRL+F (if with that one we stay in that message and don't scroll elsewhere, this I haven't tried).

daaain:
>  (if with that one we stay in that message and don't scroll elsewhere, this I haven't tried).

I don't think we can synchronise the browser search selection index so it would probably jump to the top (or wherever it was)

cboos:
> > (if with that one we stay in that message and don't scroll elsewhere, this I haven't tried).
> 
> I don't think we can synchronise the browser search selection index so it would probably jump to the top (or wherever it was)

OK, maybe we can improve on that later, but in the meantime... I really need the fixed search feature, so merging now.

PR #280, opened 2026-07-14
Escape and allowlist embedded image data URLs to prevent stored XSS

### LINKED ISSUE #277: Stored XSS: unescaped image media_type/data in embedded <img src> (user/assistant path)
## Summary

The embedded-image rendering path interpolates the image's `media_type` and base64 `data` straight into an `` attribute **without HTML-escaping, base64 validation, or a media-type allowlist**. These fields come from transcript JSON and can carry content that did not originate from the user (e.g. images returned by tools/MCP servers, fetched from web pages, or IDE/file attachments), so a crafted `media_type` can break out of the `src` attribute and inject a live `<script>` into the generated HTML — stored XSS that executes when the file is opened in a browser under `file://` (local-file read / exfiltration).

## Affected code

- `claude_code_log/image_export.py:43` builds the data URL:
  ```python
  return f"data:{image.source.media_type};base64,{image.source.data}"
  ```
- Sinks that embed it with no escaping:
  - `claude_code_log/html/assistant_formatters.py:111`
  - `claude_code_log/html/renderer.py:476`
  ```python
  return f'<img src="{src}" alt="image" class="uploaded-image" />'
  ```
- `ImageSource.media_type` and `.data` are plain unvalidated `str` (`claude_code_log/models.py:110-111`), parsed directly from JSON.

`embedded` is the default HTML image mode, so this is the normal path.

## Reproduction

A user-turn image content block with:

```
media_type = png"><script>alert(document.domain)</script>
```

renders as:

```html
<img src="data:png"><script>alert(document.domain)</script>;base64,AAAA" ... >
```

The `">` closes the `src` attribute and the `` tag; the `<script>` then executes on open. A `"` in `data` would do the same.

## Why this is an oversight, not accepted risk

The **tool-result** image path already does the right thing at `claude_code_log/html/tool_formatters.py:1479-1500`:

- allowlists `media_type` to `{image/png, image/jpeg, image/gif, image/webp}` (excludes scriptable `image/svg+xml`),
- validates the base64 with `base64.b64decode(data, validate=True)`,
- and wraps the final URL in `escape_html(data_url)`.

The user/assistant embedded-image path does none of the three.

## Suggested fix

Centralise a hardened data-URL builder (e.g. in `image_export.py`) applying the same three guards as the tool-result path, and route both `` sinks (and ideally the tool-result path) through it:

1. allowlist `media_type` (png/jpeg/gif/webp),
2. `base64.b64decode(validate=True)` the `data`,
3. `escape_html(...)` the final `src`.

Add a regression test for the `media_type = 'png"><script>...'` breakout and for the `data` quote case.

## Severity

Stored XSS in a `file://` context against a page rendered from the user's private logs. High.

evoskamp:
## What & why

The embedded-image rendering path interpolated an image's `media_type` and base64 `data` straight into an `` attribute with **no HTML-escaping, media-type allowlist, or base64 validation**. Those fields come from transcript JSON and can carry content that did not originate from the user (images returned by tools/MCP servers, fetched from web pages, IDE/file attachments), so a crafted `media_type` breaks out of the attribute and injects a live `<script>` — stored XSS that runs when the generated HTML is opened in a browser under `file://` (local-file read / exfiltration).

`embedded` is the default HTML image mode, so this is the normal path.

### Repro (before this PR)

A user-turn image block with `media_type = png"><script>alert(document.domain)</script>` rendered as:

```html
<img src="data:png"><script>alert(document.domain)</script>;base64,AAAA" ... >
```

The `">` closes the `src` attribute and the `` tag; the `<script>` executes on open. A `"` in `data` does the same.

## The fix

- New shared `_is_safe_image_source()` guard in `image_export.py`: allowlists `media_type` to `image/{png,jpeg,gif,webp}` (excludes scriptable `image/svg+xml`) and validates the base64 with `base64.b64decode(validate=True)`. Applied to **both** `embedded` and `referenced` modes; `image/svg+xml` dropped from `_get_extension`.
- Both HTML `` sinks (`html/assistant_formatters.py`, `html/renderer.py`) now wrap `src` in `escape_html`.
- This mirrors the guards already present on the **tool-result** image path (`html/tool_formatters.py`), which allowlisted, validated, and escaped — the user/assistant path had simply missed all three.

The Markdown output path (``) intentionally keeps the raw data URL (escaping would be wrong there) but still gains the upstream allowlist + base64 validation.

## Tests

Adds regression coverage in `test/test_image_export.py`:

- attribute-breakout `media_type` rejected (embedded),
- SVG rejected (embedded and referenced),
- invalid base64 rejected (embedded and referenced),
- allowlisted types still produce data URLs,
- end-to-end: the HTML `` sink emits no `"><script` breakout.

Full unit suite green; `pyright` clean on changed files.

Closes #277

## Summary by CodeRabbit

* **Security**
  * Improved image handling to prevent unsafe content from breaking HTML attributes or executing scripts.
  * Restricted supported image formats to PNG, JPEG, GIF, and WEBP.
  * Invalid or malformed image data now displays a placeholder instead of being embedded or saved.
* **Bug Fixes**
  * Added HTML escaping for image sources in rendered output.
  * Prevented unsafe image files from being written to disk.

### DIFF
--- a/claude_code_log/html/assistant_formatters.py
+++ b/claude_code_log/html/assistant_formatters.py
@@ -108,7 +108,10 @@ def format_image_content(image: ImageContent) -> str:
     src = export_image(image, mode="embedded")
     if src is None:
         return "[Image]"
-    return f'<img src="{src}" alt="image" class="uploaded-image" />'
+    # Escape the src: export_image allowlists the media type and
+    # validates the base64, but the data: URL still must not be able to
+    # break out of the attribute (issue #277).
+    return f'<img src="{escape_html(src)}" alt="image" class="uploaded-image" />'
 
 
 def format_unknown_content(content: UnknownMessage) -> str:
--- a/claude_code_log/html/renderer.py
+++ b/claude_code_log/html/renderer.py
@@ -473,7 +473,10 @@ class HtmlRenderer(Renderer):
         )
         if src is None:
             return "[Image]"
-        return f'<img src="{src}" alt="image" class="uploaded-image" />'
+        # Escape the src: export_image allowlists the media type and
+        # validates the base64, but the data: URL (embedded mode) still
+        # must not be able to break out of the attribute (issue #277).
+        return f'<img src="{escape_html(src)}" alt="image" class="uploaded-image" />'
 
     # -------------------------------------------------------------------------
     # System Content Formatters
--- a/claude_code_log/image_export.py
+++ b/claude_code_log/image_export.py
@@ -13,6 +13,47 @@ if TYPE_CHECKING:
     from .models import ImageContent
 
 
+# Image media types we are willing to emit into a data: URL or write to
+# disk. Deliberately excludes ``image/svg+xml`` — SVG can carry inline
+# ``<script>`` and event handlers, so a data:image/svg+xml URL is a
+# scriptable XSS vector when the generated page is opened under
+# ``file://``. Mirrors the allowlist already enforced on the
+# tool-result image path in ``html/tool_formatters.py`` (issue #277).
+_ALLOWED_IMAGE_MEDIA_TYPES = frozenset(
+    {
+        "image/png",
+        "image/jpeg",
+        "image/gif",
+        "image/webp",
+    }
+)
+
+
+def _is_safe_image_source(media_type: str, data: str) -> bool:
+    """Whether an image's media type and base64 data are safe to emit.
+
+    Guards the embedded (data: URL) and referenced (write-to-disk)
+    paths against two problems:
+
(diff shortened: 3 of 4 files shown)


cboos:
Thanks! Looks good & useful. Would you mind considering making `_is_safe_image_source` "public" so that it would be usable in plugins? And as such, documented in dev-docs/plugins.md (4.2).

evoskamp:
> Thanks! Looks good & useful. Would you mind considering making `_is_safe_image_source` "public" so that it would be usable in plugins? And as such, documented in dev-docs/plugins.md (4.2).

Good s suggestion. I'll deal with that this weekend.

PR #287, opened 2026-07-19
Add --depth option; make tool/high the default output level (#159)

### LINKED ISSUE #159: Make "--depth tool" the default detail level? (and deprecated "--detail)

cboos:
Closes #159.

Replaces the confusingly-named `--detail` scale with **`--depth`**, which names each level by the message-hierarchy node the output stops at (`session > user > assistant > agent > tool > hook`), defaulting to `tool`.

## What

- **`--depth session|user|assistant|agent|tool|hook`** (default `tool`), mapping onto the existing `DetailLevel` via `models.DEPTH_TO_DETAIL` (`tool`=HIGH, `hook`=FULL, `agent`=LOW, `assistant`=MINIMAL, `user`=USER_ONLY), plus a **new `session`** level (`DetailLevel.SESSION`): session structure only — session/branch headers + fork landmarks, every message body dropped (explicit branch in `_ghost_template_by_detail`, since the `visible_at` predicate keeps threshold-less built-ins like `UserTextMessage` visible at every level).
- **Default output level changes FULL → HIGH (`--depth tool`)**: a bare invocation now renders the cleaned view (no system/hook noise). Use `--depth hook` for the old everything-view.
- **`--detail full|high|low|minimal|user-only`** kept as a **deprecated** alias (removed in 2.0): still maps to the same levels, emits a stderr warning, mutually exclusive with `--depth`.
- Filenames use a **single canonical suffix per level — the `--depth` name** (`.hook/.agent/.assistant/.user/.session`), default `tool` suffix-less — regardless of which option selected the level (so `--detail low` and `--depth agent` share the `.agent` file). Old `.low`/`.user-only` filenames still *parse* (backward-compat reading).

## Design decision (approved)

Did **not** thread a dual-naming scheme to preserve old `--detail` suffix names — that needs a bool through ~20 rendering-core sites for a deprecated option, and would let cache keys/filenames diverge for identical content. Single canonical `.agent` for both spellings. (Approved by cboos + monk.)

The **library render default stays FULL** (`generate_html` etc. render everything — snapshot `.ambr` unchanged); only the **CLI/converter** default is HIGH, and the converter always threads `detail` to the renderer so filename ↔ content never diverge.

## ⚠️ Migration / release-note callouts

- **User-visible default change**: default output goes from full → cleaned.
- Stale FULL-era outputs **regenerate at the release that ships this** (the `HtmlRenderer.is_outdated` embedded-version gate fires whenever versions differ, and a released upgrade is always a distinct version). Source builds *between merge and release* re-render at the same dev version, so `is_outdated` no-ops → may need `--clear`. Narrow, dev-only, transient.
- Old-named variant files (`.low`/`.user-only`, or FULL) from prior versions persist after upgrade and list under old labels alongside the new depth-named ones — `--clear-output` to tidy.

## Tests

New `test/test_depth_option.py` (18): mapping/default/suffix/CLI-resolution/deprecation/mutual-exclusion, the `session` level (mutation-verified), and migration guards (single-file + cached-directory version-gate regen, `--detail high` ≡ default byte-equality). Updated variant-suffix/output-path/obsidian/combined-link tests to the new default + depth suffixes (old-name literals kept only in naming-agnostic parse tests). `dev-docs/application_model.md` §2.6 rewritten.

`just ci` green (unit / TUI / browser / integration / benchmark, pyright + ty).

## Summary by CodeRabbit

* **New Features**
  * Added a `--depth` option to control rendering depth across HTML, Markdown, and JSON, including session-level depth filtering.
  * Updated output variant suffixes and links to be depth-based (and session back-links accordingly).

* **Bug Fixes**
  * Preserved navigation anchors and fork back-links when content is filtered by depth.
  * Improved cache/output regeneration for default depth changes.

* **Documentation**
  * Updated CLI, rendering, plugin, and workflow docs to use depth terminology.
  * Documented `--detail` as a deprecated compatibility alias for depth.


cboos:
(Claude) Good catch, CodeRabbit — fixed in f77a760. The blanket `detail`→`depth` word-rename swept up the English idiom "implementation detail" (and, on a sweep for stale prose, "describe … in detail" / "add more detail" in two test fixtures). Restored all three — they were the noun, not the rendering depth.

While there, refreshed the remaining stale bare old-level names in code comments/docstrings/dev-docs to the new members (`FULL`→`HOOK`, `HIGH`→`TOOL`, `LOW`→`AGENT`, `MINIMAL`→`ASSISTANT`, `USER_ONLY`→`USER`). Comments only — the deprecated `--detail` option, HTML `<details>` elements, and `DETAIL_ALIASES` are untouched, and no test string literals changed (the rename stays byte-identical: snapshot `.ambr` unchanged, `just ci` green).

cboos:
(Claude) Fixed in 6c332e9 — thanks. The `detail`→`depth` rename was code-only, so dev-docs kept half-migrated prose ("HOOK detail", `--detail high` as the canonical example). Updated teammates.md / workflows.md / application_model.md / plugins.md to describe the canonical `--depth` (with "HOOK depth", `--depth tool`, `--depth agent`), keeping `--detail` solely as the deprecated-alias reference. Docs-only, no code/output change.

PR #290, opened 2026-07-23
feat: add claude-history-mcp to community extensions

sydasif:
Adds [claude-history-mcp](https://github.com/sydasif/claude-history-mcp) as a community extension in the README.

This MCP server lets Claude Code query its own session history — search messages, list sessions, retrieve transcripts, and analyze usage patterns across all your projects.

Key features:
- Reads from `~/.claude/projects/**/*.jsonl` (session transcripts) and `~/.claude/history.jsonl` (command history)
- Caches data in a local SQLite database with incremental mtime tracking
- Exposes 7 MCP tools and 2 resources via FastMCP (stdio transport)
- Natural language date filtering ("yesterday", "last week", etc.)
- Install: `claude mcp add claude-history --scope user -- uvx --from git+https://github.com/sydasif/claude-history-mcp claude-history-mcp`

## Summary by CodeRabbit

* **Documentation**
  * Updated the Community Extensions section to include `claude-history-mcp`.
  * Added details on its transcript/history input locations, local SQLite caching with incremental tracking, and the number of FastMCP tools and resources available.
  * Adjusted the related projects listing formatting and updated the TODO item formatting for GitHub Actions.

PR #292, opened 2026-07-24
feat: add public library API for external consumers

sydasif:
Adds a stable public library API (`claude_code_log.api`) so external projects like claude-history-mcp can depend on this library instead of duplicating parsing logic.

**Changes:**
- New `claude_code_log/api.py` with public functions:
  - `load_transcript()`, `load_directory_transcripts()` - core loading
  - `CacheManager`, `SessionCacheData` - cache access
  - `ensure_fresh_cache()` - cache refresh
  - `create_transcript_entry()`, `extract_text_content()`, `parse_timestamp()` - parsing utilities
  - `discover_projects()`, `find_history_file()`, `load_history_file()` - discovery
  - `TranscriptEntry` - base type

- Updated `__init__.py` to expose `api` submodule and `__version__`

**Use case:** Allows `claude-history-mcp` to import parsing/loading from here instead of maintaining duplicate code. Once merged, they'll add `claude-code-log @ git+https://github.com/daaain/claude-code-log` as a dependency and refactor their parser/loader/cache to use this API.

**Follow-up:** After this merges, I'll add a `--refresh-cache-only` CLI mode (as discussed in #290) so claude-history-mcp can trigger cache refreshes programmatically.

## Summary by CodeRabbit

* **New Features**
  * Added a public `claude_code_log.api` library API for transcript/history loading, project discovery, and cache/session helpers.
  * Updated Codex docs rendering control to use rendering depth visibility (replacing the previous detail-level setting).
  * Exposed package version metadata.
* **Database / Cache**
  * Added a `history_commands` table and bulk insertion with duplicate avoidance.
* **Tests**
  * Added regression coverage for the public API, history parsing/loading, and cache behavior.
* **Chores**
  * Updated the minimum supported `pydantic` version.


cboos:
Looks good! Now we just need regression tests so we don't accidentally break that API... 

Also note that as we're full-throttle adding support for *other* coding agents, we may want to make sure this public API supports that as well (provider parameter somewhere). So maybe #291/#293 (upcoming) first, then revisit #292.

cboos:
Tests pass, but pyright failed. You'd typically avoid that by running `just ci` before push (don't worry, I also forget to do that from time to time).

cboos:
> "Merge branch 'main' into add-library-api"

No, that's not the way you should do it: you should rather *rebase* your branch on an updated main.

Assuming you're on that merge commit in your branch, you could do:

```
git update-ref HEAD HEAD^ # "undo"
git checkout main
git pull # get latest main
git checkout add-library-api
git rebase main
git push --force
```

sydasif:
> > "Merge branch 'main' into add-library-api"
> 
> No, that's not the way you should do it: you should rather _rebase_ your branch on an updated main.
> 
> Assuming you're on that merge commit in your branch, you could do:
> 
> ```
> git update-ref HEAD HEAD^ # "undo"
> git checkout main
> git pull # get latest main
> git checkout add-library-api
> git rebase main
> git push --force
> ```

Ok, thanks for guidence

sydasif:
Done! Rebased the branch onto main and force-pushed. The merge commit has been removed.

All tests pass (2639 passed, 14 skipped) and the history is now clean and linear.

sydasif:
Thanks, I will create a new pr.

PR #298, opened 2026-07-25
Recover provenance-laundered Codex tool-result correlations

cboos:
## Recover provenance-laundered Codex tool-result correlations

Builds on the QuickJS-based analysis of Codex code-mode `exec` snippets. That
mapper runs a snippet with its tool results mocked as sentinel-bearing proxies
to learn the call → output-row structure, then maps the real paired output back
onto each call. When a snippet **reads** a result but transforms the provenance
sentinel away before emitting — `JSON.parse(r.output)` throwing on the sentinel
into a `catch`, `String(r.output).indexOf(...)` missing it — the emitted row
carries no provenance, the every-call-maps invariant fails, and the whole batch
fell back to a raw `ToolExecution` render even though the calls and arguments
were captured perfectly.

On the canonical local session this dropped **21** exec programs to the raw
fallback; corpus-wide the canonicalization rate was **97.0%**. This change
recovers the laundered class: canonical fallbacks **21 → 8** (the remainder is
a genuinely-ambiguous `ALL_TOOLS`-introspection class that correctly stays raw),
corpus **97.0% → 98.0%**.

### What is recovered

- **Single-call** — a snippet with exactly one (non-wait) tool call has no
  output-split ambiguity by construction: whatever it did, the one paired
  `function_call_output` is that call's result. It is now recovered regardless
  of what (or whether) it emitted — laundered, zero-emission, or an
  undecomposable object — as a fallback *after* the strict sentinel correlation,
  so cleanly-correlated single calls keep their richer mapping (object-key
  projection, session markers).
- **Interleaving / mixed provenance** — a loop that emits one content row per
  call in execution order is attributable by position (calls-made-so-far,
  recorded per emission) even when some rows kept the sentinel and others
  laundered it. A sentinel-bearing row must self-identify the same call its slot
  implies, or the conflict fails closed.

### What still fails closed (deliberately)

Multi-call correlation remains conservative — a forged reference to a
nonexistent call, a slot-vs-sentinel conflict, a count mismatch, an unread call,
and out-of-order emission all keep the batch on the raw fallback rather than
risk mis-attributing one call's output to another. Single-call recovery is the
only unconditional path, because with one call there is nothing to mis-attribute.

### Workflow-family exclusion

`spawn_agent` (→ Task) and Workflow calls are excluded from the single-call
whole-output fallback, so their opaque agent payloads stay on the scrubbed
`ToolExecution` path; a cleanly-correlated single `spawn_agent` still renders as
a Task. A `whole_output_fallback` marker on the batch distinguishes the two.

### Robustness

The single-call fallback batch keeps its per-call metadata tuples call-length
(an earlier iteration defaulted them empty, which made the batch consumer
`IndexError` before it even checked the call count — it would have crashed the
whole transcript render). The exec batch-adapter call is now guarded so a future
adapter fault degrades to raw rendering instead of aborting, with a contract
test that the adapter never raises on any analysis output.

### Testing

Synthetic fixtures and mutation-checkable pins for both recoveries, each
fail-closed boundary (forged-ref, slot-vs-sentinel conflict, count mismatch,
unread call, out-of-order), the Workflow gate, and the consumer never-raises
contract (unit + end-to-end render). The one documented residual — a laundered
row could be mis-attributed only via a *dead* read of its slot call, which a
real snippet never produces — is recorded in the code with a backlog note for a
uniqueness-based tightening. Full `just ci` green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Codex transcript rendering resilience by gracefully handling batch-adapter failures, emitting a warning that includes the affected call ID.
  * Enhanced JavaScript tool execution analysis with more accurate emission/correlation tracking, stronger interleaving/provenance-laundering recovery, and clearer whole-output fallback behavior for single-call scenarios.
  * Updated widened single-call classification so single tool invocations render correctly as command output (including special-case handling for workflow-like tools).
* **Tests**
  * Expanded regression and adversarial coverage for single-call recovery, whole-output fallback, forged-reference handling, and improved nested MCP transport payload retention.

### DIFF
--- a/claude_code_log/providers/codex.py
+++ b/claude_code_log/providers/codex.py
@@ -1411,7 +1411,19 @@ class CodexProvider(BaseProvider):
             ):
                 source = record.payload.get("input")
                 if isinstance(source, str):
-                    batch = adapt_codex_tool_batch(source)
+                    # Defense-in-depth: a batch-adapter bug must degrade this
+                    # exec to its raw fallback, never crash the whole transcript
+                    # render. The adapter's own contract (never raise on any
+                    # analyze output) is pinned in the tests; this guard keeps a
+                    # contract violation from being fatal in production.
+                    try:
+                        batch = adapt_codex_tool_batch(source)
+                    except Exception:
+                        logger.warning(
+                            "Codex batch adapter raised on an exec snippet; "
+                            "falling back to raw rendering"
+                        )
+                        batch = None
                     if batch is not None:
                         requests[call_id] = (
                             batch.calls,
--- a/claude_code_log/providers/codex.py
+++ b/claude_code_log/providers/codex.py
@@ -1419,9 +1419,16 @@ class CodexProvider(BaseProvider):
                     try:
                         batch = adapt_codex_tool_batch(source)
                     except Exception:
+                        # Name WHICH exec faulted (call_id) and carry WHAT broke
+                        # (exc_info): this guard exists to diagnose an adapter
+                        # contract violation, so a non-specific warning defeats
+                        # its purpose — it was exactly this path that hid a
+                        # consumer IndexError until a corpus probe surfaced it.
                         logger.warning(
-                            "Codex batch adapter raised on an exec snippet; "
-                            "falling back to raw rendering"
+                            "Codex batch adapter raised on exec snippet %s; "
+                            "falling back to raw rendering",
+                            call_id,
+                            exc_info=True,
                         )
                         batch = None
                     if batch is not None:
--- a/claude_code_log/providers/codex_quickjs.py
+++ b/claude_code_log/providers/codex_quickjs.py
@@ -81,6 +81,11 @@ class JavaScriptToolBatch:
     synthetic_results: tuple[Optional[str], ...] = ()
     output_count: int = 0
     result_object_keys: tuple[Optional[str], ...] = ()
+    # True only for the single-call whole-output fallback (a lone call whose
+    # emissions didn't correlate). Lets the adapter keep Workflow-family calls
+    # (spawn_agent → Task) on their scrubbed-opaque fallback instead of exposing
+    # the payload; a cleanly-correlated single call leaves this False.
+    whole_output_fallback: bool = False
 
 
 # --------------------------------------------------------------------------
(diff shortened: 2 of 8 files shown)


cboos:
(Claude) @coderabbitai full review

cboos:
(Claude) Thanks — all three addressed in this PR:

**`codex.py` batch-adapter warning (`call_id` + `exc_info`)** — fixed in `796fc31`. The guard already logged on an adapter fault, but non-specifically; it now names *which* exec faulted (`call_id`) and carries the underlying exception (`exc_info=True`), which is what makes a contract violation in the adapter actually diagnosable. Pinned with a `caplog` test that forces the adapter to raise and asserts the warning carries both — mutation-checked (reverting the enrichment fails the assertion).

**Unreachable `return None` in `codex_quickjs.py`** — fixed in `796fc31`; dead code removed.

**Snippet-writable `__reads` bookkeeping** — declined as a code change in this PR, and tracked instead as a backlog item (`f3e0658`, `work/codex-backlog.md`). Your aside is the important part and the item is scoped to it: `__reads` isn't special — `__records`, `__texts`, and `__errors` are the same plain `globalThis` arrays, and a snippet that can forge reads can equally forge records/texts, i.e. fabricate whole tool calls in the rendered output, which is strictly worse than influencing one attribution. Hardening `__reads` alone would shut the narrow door and leave the wide one open, while touching already-reviewed instrumentation. The right fix is uniform — all four behind a closure exposing a single non-enumerable, non-writable extraction hook — which is a mapper-wide change rather than a nitpick follow-up. The backlog entry records exactly that, with the threat model that bounds its priority: the analyzed snippet originates in the user's own transcript and the QuickJS sandbox + evaluation caps hold, so the worst outcome is misleading rendered output for the user viewing their own session — not sandbox escape or exfiltration.

@coderabbitai full review

cboos:
(Claude) You're right, and thanks — this was a real gap in the requirement itself, not just the wording. `non-enumerable`/`non-writable` seal the property binding, not the array a hook hands back; an extraction hook returning the live closure-owned arrays would let a snippet call it and push forged entries straight into them, reintroducing exactly the forgery the closure was meant to close. The backlog entry now requires the hook to return a detached snapshot — serialized, or a deep copy / frozen structure — and never a live reference (`9b4eece`).

PR #300, opened 2026-07-26
feat: add --branches main to drop rewound-and-abandoned forks

KKingdong:
## Problem

Rewinding a Claude Code session — `/rewind`, or editing an earlier prompt — does not
delete anything. It appends a message whose `parentUuid` points at an *earlier*
message, so the transcript becomes a tree and every abandoned attempt stays in the
JSONL forever.

`claude-code-log` already models this correctly: `dag.py` walks the forks and renders
each one as its own `Branch • <uuid8> • <preview>` section. That is the right default —
nothing is hidden from the reader.

For reading back a long working session, though, it is noise. Measured on a real
4 MB / 1451-line transcript of mine:

| | count |
|---|---|
| entries with a `uuid` | 1030 |
| leaves (conversation tips) | 31 |
| DAG-lines | 12 (11 of them `is_branch`) |
| abandoned entries | 55 (**3%** of the file) |
| **abandoned user prompts** | **11 of 43 (26%)** |

The byte cost is small, but one in four questions the reader scrolls past was rewound
away and never answered. Because each fork also gets a section header *and* every
message title repeats the message's own opening line as an excerpt, the same question
can appear three times before its answer.

There is no flag to filter on: Anthropic never marks dead branches — `isSidechain`
stays `False` on every node in a rewound tree (anthropics/claude-code#24471). The shape
of the graph is the only signal available.

## Solution

New CLI option, defaulting to current behaviour:

```
--branches [all|main]   [default: all]
```

`select_main_lines()` keeps, per trunk, the single root→leaf path carrying the most
messages, and truncates each ancestor at the fork point it was left at.

### Why longest path, not latest timestamp

The obvious rule — "keep the newest leaf" — is wrong on real data. In the sample
transcript the newest leaf is a stray `attachment` whose path is **12** entries; the
thread the session actually spent its time on is **975**. Selecting by recency keeps
the stub and discards the work. Newest leaves are routinely a one-line afterthought
typed after a rewind.

### Why not just filter `is_branch`

Also wrong, and worth stating because it looks correct. In the sample, the *last* real
exchange lives on `Branch • d07feb77`, not on the trunk: after a rewind the session
continues on the branch line. Dropping `is_branch` lines discards the most recent work.
The surviving path legitimately traverses 5 branch lines.

### Sidechains

Never compete. Subagent transcripts are real work branching off the trunk, not
abandoned attempts, so they are excluded from the selection and re-admitted whenever
their anchor message survives the prune.

### Where the filter runs

At the entry level, after `deduplicate_messages` and before rendering. Once the
abandoned entries are gone there is nothing left to fork, so the DAG the renderer
builds is already linear and every existing branch code path naturally emits nothing.
**No renderer changes were needed.** A prebuilt `SessionTree` (directory mode,
workflow runs) is rebuilt from the survivors with its workflow links carried across.

Provider-wholesale rendering (Codex et al.) is deliberately untouched — the
rewind-fork shape is specific to Claude Code's session format.

## Result

```
claude-code-log s.jsonl --format md --detail minimal --compact --no-recaps --branches main
```

| | `--branches all` | `--branches main` |
|---|---|---|
| size | 105 KB | 101 KB |
| lines | 1731 | 1656 |
| `Branch` sections | 11 | **0** |
| user prompts | 24 | **18** |

## Tests

`test/test_branch_pruning.py`, 7 cases:

- abandoned branch and its reply are dropped
- longest path wins over a later stray prompt
- default (`all`) still produces every fork — behaviour unchanged
- pruned result is linear (no `is_branch` lines survive re-analysis)
- ancestor is truncated *at* the fork point, keeping the fork message itself
- a transcript with no rewinds passes through identically
- entries without a `uuid` (summaries) pass through

Existing suite: no regressions. Pre-existing failures in my environment
(`test_tui.py`, `*_browser.py`, `test_index_timezone.py`) fail identically on `main`
and are missing-optional-dependency errors, not related to this change.

## Backward compatibility

Default is `--branches all`. Output is byte-identical to before unless the flag is
passed.

## Possible follow-ups (not in this PR)

- `--no-excerpt-titles` — message headings repeat the body's opening line, which is the
  other half of the duplication described above (`markdown/renderer.py`, `_excerpt()`).
- `--no-recaps` leaves an empty `#### Recap` heading + timestamp behind for one of four
  recaps in my sample; the body is removed but the section is not. Happy to file
  separately with a repro.
- `file-history-delta` is reported as an unrecognized message type.

## Summary by CodeRabbit

* **New Features**
  * Added a `--branches` option to control transcript rendering.
  * Choose between displaying all conversation branches or only the main path.
  * Main-path rendering removes abandoned rewind branches while preserving relevant conversation history.
* **Bug Fixes**
  * Improved branch selection to retain the longest surviving conversation path rather than relying on timestamps.
  * Preserved summaries and other transcript entries without identifiers during branch filtering.

### DIFF
--- a/claude_code_log/cli.py
+++ b/claude_code_log/cli.py
@@ -101,6 +101,7 @@ def _render_provider_input_file(
     compact: bool,
     no_timestamps: bool,
     no_recaps: bool,
+    branches: str,
     open_browser: bool,
 ) -> None:
     """Render a single provider session file handed in as an INPUT_PATH.
@@ -204,6 +205,7 @@ def _run_provider_wholesale(
     compact: bool,
     no_timestamps: bool,
     no_recaps: bool,
+    branches: str,
     write_combined: bool,
     write_individual: bool,
     from_date: "Optional[str]",
@@ -1034,6 +1036,20 @@ def _validate_git_link_template(template: str) -> None:
     ),
 )
 @click.option(
+    "--branches",
+    type=click.Choice(["all", "main"]),
+    default="all",
+    show_default=True,
+    help=(
+        "Which conversation branches to render. Rewinding (/rewind, or "
+        "editing an earlier prompt) forks the transcript instead of "
+        "deleting anything, so abandoned attempts stay in the JSONL and "
+        "are rendered as extra 'Branch' sections. 'main' keeps only the "
+        "longest root-to-leaf path per trunk, dropping the attempts that "
+        "were rewound away."
+    ),
+)
+@click.option(
     "--debug",
     is_flag=True,
     default=False,
@@ -1069,6 +1085,7 @@ def main(
     git_link: Optional[str],
     no_timestamps: bool,
     no_recaps: bool,
+    branches: str,
     debug: bool,
 ) -> None:
     """Convert Claude transcript JSONL files to HTML or Markdown.
@@ -1384,6 +1401,7 @@ def main(
                     compact,
                     no_timestamps,
                     no_recaps,
+                    branches,
                     write_combined,
                     write_individual,
                     from_date,
@@ -1411,6 +1429,7 @@ def main(
                     compact,
                     no_timestamps,
                     no_recaps,
+                    branches,
                     open_browser,
                 )
                 return
@@ -1659,6 +1678,7 @@ def main(
                         compact=compact,
                         no_timestamps=no_timestamps,
                         no_recaps=no_recaps,
+                        branches=branches,
                     ),
                 )
                 return
@@ -1674,6 +1694,7 @@ def main(
                 compact=compact,
                 no_timestamps=no_timestamps,
                 no_recaps=no_recaps,
+                branches=branches,
             )
             click.echo(f"Successfully exported session to {output_path}")
             if open_browser:
@@ -1735,6 +1756,7 @@ def main(
                 write_combined=write_combined,
                 no_timestamps=no_timestamps,
                 no_recaps=no_recaps,
+                branches=branches,
                 jobs=jobs,
             )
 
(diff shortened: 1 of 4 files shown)


cboos:
Hello! Thanks for your contribution. This looks useful indeed, though I have some qualms about selecting the "main" branch. What if the "longest" is not what you want to isolate? Say you go down a long dead end, come back, start a different, shorter, and more successful route? One possibility would be to give a short message uuid (the [uuid] button can show them), `--branches uuid` (or `--branches uuid, uuid, ...`, as you already have a plural here).

The other thing is that we take the CodeRabbit suggestions seriously, so you should address them or refute them if you find them undeserved (but here they seem justified).

KKingdong:
> Hello! Thanks for your contribution. This looks useful indeed, though I have some qualms about selecting the "main" branch. What if the "longest" is not what you want to isolate? Say you go down a long dead end, come back, start a different, shorter, and more successful route? One possibility would be to give a short message uuid (the [uuid] button can show them), `--branches uuid` (or `--branches uuid, uuid, ...`, as you already have a plural here).
> 
> The other thing is that we take the CodeRabbit suggestions seriously, so you should address them or refute them if you find them undeserved (but here they seem justified).

I agree with your opinion

cboos:
One more suggestion: instead of `--branches` (plural), have claude-code-log behave as of today when no `--branch` option is given, behave as with your patch (select "longest") when `--branch main` is given, and filter in all the branches containing at least one of uuid1, uuid2, ...  when given `--branch uuid1 --branch uuid2 ...` .

PR #301, opened 2026-07-26
Render <system-reminder> blocks in user messages (#275)

### LINKED ISSUE #275: Support system reminders in user messages
Image

```xml
<system-reminder> The session's working directory has changed to .../main (via /cd). The environment block at the start of this conversation still names the previous directory — that information is stale. All tool calls and relative paths now resolve from .../main. </system-reminder>
```

This is not necessarily the *whole* message. If you `/cd` to a directory that has a CLAUDE.md, its content will come after that system reminder.

cboos:
Closes #275.

Claude Code injects `<system-reminder>` blocks into user messages — for
example the working-directory-changed notice from a `/cd`, which may be
followed in the *same* message by the content of a newly-in-scope `CLAUDE.md`.
Previously the raw tags rendered verbatim in the transcript (and leaked onto
the index preview), mixing harness scaffolding into the user's own text.

This peels each `<system-reminder>` out of the user message as an **annotation**
and renders it distinctly (🤖, italic), while the remaining real user content
continues to render as normal user text. It follows the existing
IDE-notification precedent exactly: the reminder becomes an ordered content
item on the user message, so it stays message-class `user` — no new message
class, timeline branch, or filter toggle, and thus no parity surface to keep in
sync.

- `models`: `SystemReminderContent` dataclass, added to the `UserTextMessage`
  items union.
- `factories`: `extract_system_reminder_content` peels the block(s) in the
  user-text fallback (mirroring the IDE-notification extractor); returns the
  text unchanged when there is no reminder, so reminder-free messages are
  byte-identical.
- HTML: `format_system_reminder_content` renders the reminder (escaped) with a
  new `.user .system-reminder` selector; the existing tool-result reminder
  path (`.read-tool-result .system-reminder`) is untouched.
- Markdown: renders the reminder as a blockquote rather than dropping it
  (IDE notifications are dropped in Markdown; a reminder can carry the only
  copy of following content, so dropping it would lose content).
- Session preview: a reminder-only message is no longer used as a session
  starter, and previews strip reminders first so the index shows real content
  instead of raw tags.

Tests pin extraction, item structure, HTML render + escaping, the
reminder-only edge, the session-starter/preview behaviour, and the Markdown
render-not-drop path (mutation-verified). Snapshot deltas are limited to the
new CSS rule; reminder-free messages are byte-identical.

## Summary by CodeRabbit

* **New Features**
  * Added support for `<system-reminder>...</system-reminder>` blocks inside user messages, preserving their exact position among surrounding text.
  * Reminders now render as styled annotations in HTML and as blockquotes in Markdown, including multi-reminder ordering.
  * Session previews now strip reminder markup, and reminder-only messages no longer act as session starters.
* **Bug Fixes**
  * Reminder text is safely escaped and raw reminder tags aren’t displayed.
* **Tests**
  * Added end-to-end coverage for parsing, rendering, escaping, and session-preview behavior, including HTML snapshot regression tests.


cboos:
(Claude) @coderabbitai full review

cboos:
(Claude) Thanks for the review — all six addressed in d0432cd:

- **Reminder ordering (Major):** now positional. Each `<system-reminder>` is emitted at its original position instead of prepending one aggregate, with intervening text segments still flowing through IDE-notification extraction and image handling. Measured across the sample transcripts, ~14% of reminder-bearing user messages carry text *before* the reminder (including the "session continued" shape) and some carry multiple reminders, so prepending reordered real content. The now-unused aggregate extractor is removed. Pinned with the three real shapes (text-before, text-before+after, multiple reminders), mutation-checked against reverting to prepend.
- **Markdown HTML protection (Major):** each reminder now runs through `_protect_html_tags()` before blockquoting, matching the `TextContent` path — a reminder containing raw HTML can no longer survive as live HTML in generated Markdown. Pinned and mutation-checked.
- **CSS `white-space: pre-wrap`:** added to `.user .system-reminder` so multi-line reminders keep their newlines.
- **Test assertions:** reject both raw and escaped opening/closing tags on the HTML, Markdown, and preview paths; assert each reminder renders as a `<div>` rather than only counting fragments.

Snapshot delta is limited to the new CSS rule.

(Claude) @coderabbitai full review

cboos:
(Claude) Both addressed in 6fc61f0:

- **Preview weld:** stripping a `<system-reminder>` now replaces the block *and* the whitespace hugging it with a single space, so `before<reminder>x</reminder>after` previews as `before after` (not `beforeafter`), and a mid-text reminder leaves no double gap. Pinned on the text-before+after shape and mutation-checked against the old empty substitution.
- **Test:** the multiple-reminders item-structure test now also asserts the first and last text segments survive in place.

Also added in a87da2e: an end-to-end snapshot fixture covering the positional shapes (text-before + multi-line reminder + text-after, and reminder-only).

(Claude) @coderabbitai full review

PR #303, opened 2026-07-27
Place UUID-less queue-op/steering entries in chronological order

### LINKED ISSUE #295: `--session-id` flavor is broken
I'm not sure what's going on with `--session-id`. The usual way I render a session works:

```
$ uv run claude-code-log .../8116e1a9-c108-4596-9afe-830feea77e2e.jsonl -o .examples/session-8116e1a9-c108-4596-9afe-830feea77e2e.html
Processing .../8116e1a9-c108-4596-9afe-830feea77e2e.jsonl...
Line 12680 of .../8116e1a9-c108-4596-9afe-830feea77e2e.jsonl: unrecognized message type 'file-history-delta' - skipping
Loading agent file .../8116e1a9-c108-4596-9afe-830feea77e2e/subagents/agent-a50d5e8fe80fe4b3f.jsonl...
Loading agent file .../8116e1a9-c108-4596-9afe-830feea77e2e/subagents/agent-a94565a4876eb43b2.jsonl...
WARNING: steering suppression imbalance in session 8116e1a9-c108-4596-9afe-830feea77e2e under version 2.1.218: a 'remove' op has no matching queued_command card — rendering it as legacy steering (content preserved). The remove↔queued_command 1:1 pairing appears violated (seen in real archives).
Successfully converted .../8116e1a9-c108-4596-9afe-830feea77e2e.jsonl to .examples/session-8116e1a9-c108-4596-9afe-830feea77e2e.html
```

The HTML looks as it should.

But with `--session-id`:

```
$ uv run claude-code-log --session-id 8116e1a9-c108-4596-9afe-830feea77e2e -o .examples/session-8116e1a9-c108-4596-9afe-830feea77e2e.session-id.html
Loading .../d8aad10f-22b6-4798-8019-527182192864.jsonl from cache...
Loading .../97ac42fd-e5ba-4739-b53d-7085140fa4a5.jsonl from cache...
Loading .../8116e1a9-c108-4596-9afe-830feea77e2e.jsonl from cache...
WARNING: steering suppression imbalance in session 8116e1a9-c108-4596-9afe-830feea77e2e under version 2.1.218: a 'remove' op has no matching queued_command card — rendering it as legacy steering (content preserved). The remove↔queued_command 1:1 pairing appears violated (seen in real archives).
Successfully exported session to .examples/session-8116e1a9-c108-4596-9afe-830feea77e2e.session-id.html
```

I have a bunch of *older* messages after the last legit Assistant message:

Image

Ah... and those seem not to have a UUID...

Also, the cache doesn't seem to be the culprit, same defective HTML when using `--no-cache`:

```
$ uv run claude-code-log --no-cache --session-id 8116e1a9-c108-4596-9afe-830feea77e2e -o .examples/session-8116e1a9-c108-4596-9afe-830feea77e2e.session-id.no-cache.html
Processing .../d8aad10f-22b6-4798-8019-527182192864.jsonl...
Processing .../97ac42fd-e5ba-4739-b53d-7085140fa4a5.jsonl...
Loading agent file .../97ac42fd-e5ba-4739-b53d-7085140fa4a5/subagents/agent-a7b030d70ad40cf52.jsonl...
Processing .../8116e1a9-c108-4596-9afe-830feea77e2e.jsonl...
Line 12680 of .../8116e1a9-c108-4596-9afe-830feea77e2e.jsonl: unrecognized message type 'file-history-delta' - skipping
Loading agent file .../8116e1a9-c108-4596-9afe-830feea77e2e/subagents/agent-a50d5e8fe80fe4b3f.jsonl...
Loading agent file .../8116e1a9-c108-4596-9afe-830feea77e2e/subagents/agent-a94565a4876eb43b2.jsonl...
WARNING: steering suppression imbalance in session 8116e1a9-c108-4596-9afe-830feea77e2e under version 2.1.218: a 'remove' op has no matching queued_command card — rendering it as legacy steering (content preserved). The remove↔queued_command 1:1 pairing appears violated (seen in real archives).
Successfully exported session to .examples/session-8116e1a9-c108-4596-9afe-830feea77e2e.session-id.no-cache.html
```

Well, looks like here it read two more sessions, d8aad10f-22b6-4798-8019-527182192864.jsonl and 97ac42fd-e5ba-4739-b53d-7085140fa4a5.jsonl, another issue.

cboos:
Fixes #295.

## The bug

Directory-mode loading (`load_directory_transcripts`) builds the DAG from entries
that carry a `uuid`, then re-adds the UUID-less ones — summaries, ai-titles and
**queue-operations** — by appending them at the end of the transcript.

A queue-operation `remove` is a *mid-conversation* steering interjection. Appended
at the end, it renders **after the last assistant message**, at the very bottom of
the page, detached from the moment it belongs to. On a real archive that is
hundreds of steering cards piled below the end of the conversation.

Two things follow that aren't obvious from the issue:

- **It is not specific to `--session-id`.** That's just where it was noticed. The
  same load path serves the default combined directory render, so
  `combined_transcripts.html` had the identical defect.
- **Single-file renders were never affected.** `load_transcript` keeps file order,
  so the steering entries were always in position there — which is why the two
  surfaces disagreed.

There is a second-order effect. The renderer pairs a `remove` op with its
`queued_command` card and suppresses the redundant copy. With the op exiled to the
end of the list that pairing fails, so the entry rendered *twice*: once in place,
once again in the trailing dump.

## The fix

Split the re-added entries by whether they carry a timestamp:

- **queue-operations** (`timestamp` + `sessionId`, no `uuid`) are spliced into
  chronological position within their own session;
- **summaries / ai-titles** (no timestamp) stay appended — they are session
  metadata the renderer consumes out of band, never trailing conversation.

A session's DAG entries are neither contiguous in the traversal (forks splice other
sessions between them) nor timestamp-monotonic (fork branches reorder), so a
queue-op cannot be placed against a block — it is anchored to a specific **entry**:
the same-session DAG entry with the greatest `(timestamp, index)` whose timestamp
is `<=` its own, inserted immediately after it. An op predating all of its
session's entries goes before the first one.

**Timestamp ties are the common path, not an edge case** — about 30% of queue-ops
in a real archive share a timestamp with a DAG entry. The per-session list is
sorted by `(timestamp, index)` and searched with `bisect_right` against a sentinel
index, so the anchor is the *highest-indexed* entry sharing that timestamp: the op
lands after all of them. Multiple ops resolving to the same anchor keep file order.
Deterministic; nothing relies on unsorted iteration.

## Measured

On a real 19,839-entry project directory, comparing the same load with and without
the splice:

| | before | after |
|---|---|---|
| entries loaded | 19,839 | 19,839 — a pure permutation |
| rendered `steering` messages | 313 | 229 |
| messages after the last assistant | 98 | 9 |

The 84 steering messages that disappear are the relocated ops finally pairing with
their `queued_command` cards and being suppressed as duplicates; the handful that
stay unpaired now render *in position* instead of at the end. None of the nine
remaining trailing messages is a steering card — they are genuinely recent entries
(hook attachments, a user message, two tool pairs). One task-notification that
rendered twice collapses to a single copy, content intact.

Pagination is unaffected by construction: page assignment keys only on per-session
`message_count` and `first_timestamp`, both invariant under reordering. Verified.

## Tests

`test/test_session_id_ordering.py` pins the chronological load position, the
`--session-id` export, the combined directory render, the timestamp-tie collision,
and the intent that a session's `last_timestamp` reflects its chronologically-last
entry. All mutation-checked: neutralising the splice fails every one of them.

A separate test pins the per-session **sort** specifically. It is load-bearing for
forked sessions but invisible to monotonic fixtures — deleting it left the rest of
the suite green — so the fixture is deliberately non-monotonic: list order
`07:00:10, 07:00:30, 07:00:20` with a queue-op at `07:00:25`, which must anchor to
its true chronological predecessor. Three entries is the minimum that
discriminates; a two-entry inversion still bisects to the right slot unsorted.

**The snapshot fixture is the point of the second commit.** This whole bug class
was invisible to the snapshot suite: no fixture contained a queue-operation entry,
and every snapshot test but one renders through the untouched single-file path. A
fully green regression suite said nothing about the surface that broke.
`session_id_steering_order.jsonl` closes that — a minimal synthetic conversation
with one mid-conversation `remove`, rendered through the dir/combined path, with
the steering card captured *in position*. Any future reordering of that composition
moves the snapshot and has to be explained.

## Follow-ups (noted in `work/rendering-next.md`, not addressed here)

- A session's `last_timestamp` is derived from list order rather than from the
  maximum timestamp — correct now that the list is chronological, but fragile.
- `--session-id` loads the whole project directory before filtering. That is
  deliberate (prefix resolution, archived-session fallback) and sibling files
  contribute nothing to the output, so it is wasteful rather than incorrect.

## Summary by CodeRabbit

* **Bug Fixes**
  * Steering/queue-operation entries are now inserted into the correct chronological position within each session’s transcript.
  * Same-timestamp tie-breaking and non-monotonic session scenarios now render in the expected order.
  * Session “last timestamp” and generated HTML output now reflect the true chronological endpoint.

* **Tests**
  * Added end-to-end and directory/session-specific coverage for chronological steering placement and edge cases.
  * Updated the HTML snapshot to match the corrected rendering.

* **Documentation**
  * Added notes on future improvements for timestamp computation and session-id performance behavior.


cboos:
(Claude) Thanks — one of the three is a real gap and is now fixed in 4af11eb; the other two don't reproduce.

**Added coverage for the unanchored branches** (the nitpick on `test_session_id_ordering.py`). You were right that `pos < 0` had no direct pin, and the same applied to the anchorless fallback next to it, so both are covered now:

- a queue-op predating every entry of its own session must land before that session's *first* entry;
- a queue-op whose session contributes no DAG entries at all must still be emitted rather than dropped.

The first fixture is deliberately multi-session and leads with the *other* session. That's what makes it discriminating: in a single-session fixture, "before this session's first entry" and "at the front of the list" are both index 0, so the test would pass under either implementation. Here they're positions 1 and 0. Each pin was mutation-checked and fails only its own test — front-of-list placement, append-at-end, and dropping the anchorless op respectively.

**The `ruff format` finding doesn't reproduce.** `uv run ruff format --check .` reports `229 files already formatted` (exit 0) on this branch, and `just ci` runs `ruff format` as its first step, so a formatting drift couldn't survive it.

**On the two "run pyright / just ci before pushing" comments:** those checks were run — `just ci` exits 0 (pyright `0 errors, 0 warnings`, `ty check` clean, full suite green), and CI here is green across all ten OS/Python matrix jobs. The analysis snippet in that comment shows the tooling wasn't available in the review sandbox (251 bytes of output), which I think is where the request came from.

PR #304, opened 2026-07-28
Fail fast on --snapshot-update under xdist parallelism

cboos:
## The footgun

`pyproject.toml` sets `addopts = -n auto --dist=worksteal`, so **every** pytest
invocation is parallel by default. Combine that with `--snapshot-update` and
syrupy's workers race each other writing the shared `.ambr` files.

One observed run silently **truncated ~6000 lines** out of
`test_snapshot_html.ambr`. The file was left structurally broken — and still
passed on the next read, because the assertions whose snapshots had vanished
simply had nothing left to compare against. A second parallel update, with
stale `__pycache__` in play, made an *untouched* fixture appear to regenerate
with fold-bar/children markup it had never carried; that one's mechanism was
never pinned down, but it did not survive a serial re-run with purged
bytecode.

What makes this worth a guard rather than a paragraph in the docs is that
**the unsafe combination is the default**. Nobody has to do anything unusual
to reach it — `uv run pytest --snapshot-update` is already the broken
invocation, and the damage is silent in both directions: nothing fails at
write time, and nothing fails on the next read either.

## The guard

`pytest_configure` in `test/conftest.py` raises `pytest.UsageError` when
`--snapshot-update` is combined with more than one xdist worker, pointing at
the serial form:

```
ERROR: Refusing to run --snapshot-update under pytest-xdist with 8 workers:
parallel writes to the shared .ambr snapshot files race and can silently
truncate snapshot content (one observed run dropped ~6000 lines), leaving a
broken file that still passes on the next read. Regenerate serially instead:

    uv run pytest <targets> -n0 --snapshot-update
```

The decision is a pure function —
`_snapshot_update_xdist_conflict(is_worker, update_snapshots, workers)` — so
it is unit-testable without spawning pytest, and `pytest_configure` is a thin
wrapper around it.

Three things it deliberately does **not** block:

- `-n0` (`workers == 0`) and `-n1` — single-writer updates are serial and safe;
- parallel runs that aren't updating (`update_snapshots` false) — they never
  write, so **CI is untouched**;
- the xdist *workers* themselves (`is_worker` true) — only the controller
  gates, before workers spawn.

Two implementation details worth knowing if you touch this: by
`pytest_configure` time `config.option.numprocesses` is already an `int`
(`-n auto` → 8, `-n0` → 0, no flag → `None`), so there's no `"auto"` string to
parse; and syrupy's write flag is `update_snapshots`, while
`--snapshot-warn-unused` is report-only and never writes.

## Tests

`test/test_snapshot_update_guard.py` covers both layers:

- four pure-function tests pinning each arm of the decision — fires on
  parallel update, silent at `-n0`/`-n1`, silent without `--snapshot-update`,
  silent on a worker;
- three **execution** tests that run real pytest subprocesses against a
  temporary project whose `conftest.py` re-exports the real hook, asserting
  the actual exit status and stderr rather than a mocked config object.

The subprocess tests matter here because the guard's whole job is to abort a
real invocation before it does damage; a test that only calls the predicate
would pass just as happily if the hook were never wired up.

## Verified against current `main`

The guard now shares a repository with snapshot fixtures that landed after it
was written, so all three paths were **executed**, not reasoned about:

| invocation | result |
|---|---|
| `--snapshot-update -n auto` | guard fires, exit 4, message points at `-n0` |
| ordinary parallel run, no update | silent — 10 passed, exit 0 |
| `-n0 --snapshot-update` | 10 snapshots passed, 0 generated, `.ambr` byte-identical |

The third is the one that earns its place: a guard whose escape hatch has
never been run is a guard that redirects people into an untested path. Here
the recommended path regenerates the current fixtures to themselves, leaving
the snapshot file unchanged.

## CONTRIBUTING

The guard prevents the mistake going forward, but you can still meet a
suspicious `.ambr` diff — reviewing a PR that carries snapshot changes, or
reading a historical diff from before the guard existed. So the docs also
cover **recognising** it: check at the block level (compare the set of
snapshot names, diff each block's content) rather than trusting a raw `-N`,
because inserting a snapshot or changing embedded CSS realigns shared
boilerplate and can show hundreds of "deletions" with zero content lost. A
real `-275` in this repo was exactly that benign case. Confirm either way with
a read-only `-n0` run after purging stale bytecode: if every snapshot passes,
the committed file matches what the code renders.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented snapshot updates from running with multiple parallel workers, helping avoid corrupted or unreliable snapshot files.
  * Added a fail-fast guard with a detailed message directing snapshot updates to run serially with `-n0`.
  * Confirmed that ordinary parallel test runs remain unaffected.

* **Documentation**
  * Expanded snapshot-testing guidance with troubleshooting steps for suspicious snapshot diffs and a safe verification workflow.

* **Tests**
  * Added coverage for controller vs worker behavior and end-to-end runs proving parallel update refusal, serial success, and non-updating parallel acceptance.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,17 +125,58 @@ uv run pytest test/test_snapshot_html.py -n0 --snapshot-update
 ```
 
 > **`--snapshot-update` must run serially (`-n0`) — a guard now enforces
-> this.** Syrupy and pytest-xdist race when writing snapshot files in
-> parallel, and it corrupts `.ambr` files in more than one way: once ~6000
-> lines were silently *truncated*, and separately a run *deleted and
-> rewrote* entries in a fixture that wasn't touched — each time leaving a
-> structurally-broken file that still passed on the next read (the second
-> instance also produced a false "the suite has non-isolated rendering
-> state" conclusion). Because `pyproject.toml` defaults to `-n auto`, a
-> `conftest.py` guard (`pytest_configure`) now fails fast if
-> `--snapshot-update` is combined with more than one xdist worker, pointing
-> you at `-n0`. Ordinary parallel test runs (no update) are unaffected, so
-> CI is untouched.
+> this.** Syrupy and pytest-xdist misbehave when writing the shared `.ambr`
+> files in parallel, on two observed occasions. Once, a raced update
+> silently *truncated* ~6000 lines, leaving a structurally-broken file that
+> still passed on the next read — a confirmed corruption. Separately, a
+> parallel `--snapshot-update` with a stale `__pycache__` produced a diff in
+> which an *untouched* fixture appeared to regenerate with structure it had
+> never carried (fold-bar / children markup); re-run serially with a purged
+> cache, the same operation was cleanly additive and that structure did not
+> appear. The mechanism of the second case isn't pinned down (the large
+> deletion count first quoted for it turned out to be alignment noise — see
+> "Recognising the race" below), but an operation that makes an untouched
+> fixture look different is dangerous regardless, and "vanished under `-n0`"
+> is the reproducible part. Because `pyproject.toml` defaults to `-n auto`,
+> this unsafe combination is the *default*, so a `conftest.py` guard
+> (`pytest_configure`) now fails fast when `--snapshot-update` is combined
+> with more than one xdist worker, pointing you at `-n0`. Ordinary parallel
+> runs (no update) are unaffected — CI is untouched.
+
+**Recognising the race in a diff.** The guard prevents the mistake going
+forward, but you may still meet a suspicious `.ambr` diff — reviewing a PR
+that carries snapshot changes, or reading a historical diff from before the
+guard existed. The rule of thumb:
+
+> Any negative in an `.ambr` diff is the race's signature — a purely
+> additive fixture is `+N/-0`; deletions mean either an intentional content
+> change you can name, or the race.
+
+To tell which, check at the **block level**, not the raw git diff: compare
+the set of snapshot names (a race removes or rewrites blocks you didn't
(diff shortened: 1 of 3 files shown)


cboos:
(Claude) Both findings are real and are fixed in 3310085.

**Bound the child pytest run.** Confirmed — `_run_pytest` called `subprocess.run(...)` with `capture_output=True` and no `timeout=`, and one of the three execution tests spawns real xdist workers, so an indefinite wait was a plausible failure rather than a theoretical one. On CI that isn't a slow test, it's a stuck job burning a runner until the workflow-level timeout kills it, with nothing saying which test was wedged.

Bounded at 300s — generous enough that a slow machine never trips it, since the point is to fail *by name* rather than to enforce a budget — with `subprocess.TimeoutExpired` converted into a `pytest.fail` naming the invocation that hung. I verified the path is reachable rather than assuming it: forcing the bound to 1ms produces the named failure, and restoring it returns 7/7 green. (Noting the irony you'd be entitled to point out: a test for a guard against silent CI damage could itself silently damage CI.)

**The rule of thumb contradicted its own next paragraph.** Also confirmed, and worse than a wording nit because the two sit three lines apart. It read *"deletions mean either an intentional content change you can name, or the race"* — an explicit enumeration of two — and the following paragraph introduced *"a third case, benign, that is neither a content change nor the race"*. So a passage whose entire purpose is teaching people not to over-read a negative diff over-read one in its own rule.

It now says a negative is a **signal to investigate, not a verdict**, and enumerates three cases: a nameable content change, a benign realignment of shared boilerplate, or the race. That makes the `-275` a worked example of the rule rather than an exception to it — which is what it always was.

`just ci` exit 0 (pyright 0 errors, ty clean).

PR #306, opened 2026-07-29
Resolve [Image #N] placeholders from the recorded paste ids

cboos:
A pasted image leaves an `[Image #N]` placeholder in the message text and its
bytes in a separate content block. The renderer inferred which block a
placeholder meant from **position** — `[Image #3]` was the third block — but
the transcript records the association explicitly, in a field we never read.

`imagePasteIds` runs parallel to the image blocks, so `[Image #N]` is the block
at `imagePasteIds.index(N)`.

**N is a paste counter, not a position.** It resets when the CLI restarts
inside a session that outlives it, and it increments on delete-and-repaste —
hence runs like `[4, 6, 8]`, numbers far above the block count, and the same N
meaning different images at different points in one session.

## What it was costing

Across the 168 placeholders in the 110 image-bearing messages of one real
archive (measured 2026-07-28):

| | |
|---|---|
| left as literal text, image detached | 126 |
| resolved to the correct block | 34 |
| **resolved to the WRONG block** | **8**, across 5 messages |

The last row is the one with nothing in the output to mark it: a different
screenshot renders where the placeholder sat. After the change all 168 resolve,
85 messages render differently, and none disagrees with the recorded ids.

## The fix

One mapping serves both halves of the rendering: the inliner takes blocks by
number, and the caller appends whichever were not taken. A block can therefore
never be inlined *and* appended, nor dropped by a pass that never rendered it —
which is what a placeholder buried in a `<system-reminder>` or an
IDE-notification prefix used to cause.

Where the association cannot be established, the placeholder renders literally,
the block stays detached, and a warning names the condition: a paste id we do
not have, lists that are not parallel, a repeated id, a malformed field. A
visible gap beats a confident substitution. The model field is deliberately
untyped so a malformed value reaches that report rather than failing validation
and costing the whole entry.

## Old transcripts, where nothing was recorded

Position is the only evidence there, and it is used only when the numbering is
`1..k` with no gaps and no more numbers than blocks.

**Why that condition** — stated carefully, because the obvious reason is wrong.
It is *not* that a reset counter cannot produce `1..k`; one that has just reset
produces exactly that. It is that on `1..k` **the recorded and positional
readings coincide**, so the renderer never has to know which regime produced the
numbering: paste ids are distinct, `>= 1` and recorded in ascending block order,
so a list containing `1..k` holds those values in its first `k` slots, making
`index(N)` and `N-1` the same answer.

That rests on the ascending premise, which is **measured, not guaranteed by the
format**: 180/180 messages carrying the field have strictly ascending ids, no
duplicates, none below 1. Scored the other way — pretend the field is absent and
run the fallback against the recorded ids as ground truth — it fires on 24 of the
110 placeholder-bearing messages, refuses the other 86, and is correct on all 24.
The era where the fallback actually applies has no ground truth by construction,
which is the limit of the argument.

**How old is "old" — and what the sampling does *not* establish.** In the same
archive, every image-bearing message containing an `[Image #N]` placeholder
carried the field, across versions 2.1.84–2.1.220. The oldest records lacking it
are twenty at 2.0.73–2.0.74, none of which contains a placeholder. That does
**not** pin an introduction version: both shapes occur *within* 2.0.74, and the
archive's oldest image-bearing record is 2.0.73, so the floor is a sampling
floor rather than a fact about the harness. The repo's own fixtures at 1.0.31,
1.0.43 and 1.0.128 do carry placeholders with no field, which is what keeps the
positional path live rather than historical.

The fallback is also strictly narrower than what the renderer did before: it
substituted positionally whenever `N <= len(images)`, unconditionally. So this
can only render *fewer* unverifiable substitutions than its predecessor, never
more.

## Steering cards carry their own paste ids

A steering delivery that carries an image records `imagePasteIds` in its
`queued_command` attachment rather than at entry top level — the attachment is
its own carrier. Image-bearing steering cards only began rendering at all once
the list-shaped prompt was accepted, so this completes that: 19 such cards in
the same archive, whose placeholders now resolve from the recorded ids instead
of from position.

Nothing needed adding to the model for that path: the attachment payload is
already a passthrough dict, so the field arrives intact and the only way to lose
it is to forget the argument at the call site. That is what the new test pins,
with deliberately non-contiguous ids (`[4, 6]`) — contiguous ids make the two
readings agree, so such a test would pass with the argument dropped.

## Tests

Every guard is mutation-checked: neutralise it and confirm the intended test
reddens and no other. That covers the id lookup, the parallel-length and
duplicate-id checks, the bool-is-not-an-int clause, the unknown-id refusal, both
legacy-fallback clauses, the consume-on-inline rule, the model field, and the
call sites. Fixtures are synthetic; one uses a real 1×1 PNG so the resolved
block is checked to reach the page as a usable `data:` URL rather than as a
correct index.

Zero snapshot churn, and structurally so: no fixture data file carries
`imagePasteIds`, so no existing snapshot can reach the recorded-id path, and the
three legacy fixtures that do reach the resolver all satisfy the positional
condition and render identically.

Corpus figures measured 2026-07-28.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rendering of `"[Image `#N`]"` placeholders by honoring per-message image paste IDs, including non-sequential and out-of-order references.
  * Fail-closed handling for missing, malformed, duplicate, or mismatched paste-id data to prevent incorrect substitutions or dropped images.
  * Preserved unresolved placeholders and unreferenced image blocks without unintended reordering.
  * Kept legacy positional fallback only when it’s safe/compatible.

* **Documentation**
  * Added guidance on placeholder-to-image mapping, fallback rules, and relevant edge cases.

* **Tests**
  * Added extensive tests covering resolution, failure scenarios, steering queued commands, and JSONL/transcript parsing integration.

### DIFF
--- a/claude_code_log/factories/attachment_factory.py
+++ b/claude_code_log/factories/attachment_factory.py
@@ -217,7 +217,20 @@ def _create_queued_command_message(
 
     meta = create_meta(transcript)
     result = create_user_message(
-        meta, prompt.items, extract_text_content(prompt.items), is_slash_command=False
+        meta,
+        prompt.items,
+        extract_text_content(prompt.items),
+        is_slash_command=False,
+        # A steering delivery carrying images records its own paste ids, in
+        # the attachment payload rather than at entry top level — the
+        # ``queued_command`` is the carrier here, not the user entry. Nothing
+        # declares the field on the model because ``attachment`` is already a
+        # ``dict[str, Any]`` passthrough, so forgetting *this argument* is the
+        # only way the ids get lost. Without them the ``[Image #N]``
+        # placeholders fall back to the positional reading, which is wrong
+        # wherever the paste counter is not 1..k (see
+        # ``user_factory._image_reference_mapping``).
+        image_paste_ids=payload.get("imagePasteIds"),
     )
     # Defensive only: create_user_message never returns None for a non-empty
     # chunk (it returns None solely on an empty content_list / the empty
--- a/claude_code_log/factories/user_factory.py
+++ b/claude_code_log/factories/user_factory.py
@@ -716,9 +716,16 @@ def _image_reference_mapping(
     """
     if paste_ids is None:
         # Old transcripts recorded no association, so the positional reading
-        # is the only one available. It is trustworthy exactly when the
-        # numbering is 1..k with no gaps and no more numbers than blocks —
-        # the shape a paste counter that has reset or skipped cannot produce.
+        # is the only one available. Use it only for 1..k with no gaps and no
+        # more numbers than blocks — not because a paste counter cannot
+        # produce that shape (one that has just reset produces exactly it),
+        # but because on that shape the two readings COINCIDE and we do not
+        # have to know which regime we are in: paste ids are distinct, >= 1
+        # and recorded in ascending block order, so a list containing 1..k
+        # must carry them in its first k slots, making index(N) == N-1. The
+        # ascending premise is measured, not guaranteed — see
+        # dev-docs/messages.md. A gap or a number above the block count
+        # breaks the coincidence, and those fail closed.
         if not numbers:
             return {}, []
         if numbers == list(range(1, len(numbers) + 1)) and len(numbers) <= len(images):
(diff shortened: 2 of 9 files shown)


cboos:
@coderabbitai review

Issue #151, opened 2026-05-10
Support Obsidian-friendly output


cboos:
> | Bullet | Reading |
> |-|-|
> | "obsidian plugin" | Integrate into Obsidian vault so transcripts become first-class notes that can be cross-referenced from other notes. |
> 
> I'm thinking of adding some additional flexibility to improve Obsidian support (or similar Markdown-based KM tools).
> 
> - With `--output`, the Markdown files are generated under a target directory, but under a "flat" structure, mirroring the one in `~/.claude/projects` (already there)
> - With the `--expand-paths` flag, a target project could be "projected back" onto its real path: `-home-joe-project-A` => `<output>/home/joe/project/A`.
> - With `--filter-path`, only the matching projects would be selected, and then the corresponding path would be truncated from the output.
> 
> All the three together:
> 
>     claude-code-log --output ~/Documents/Obsidian/ClaudeProjects --expand-paths --filter-path /home/joe --format md --detail low --compact
> 
> Would result in rendered transcript being generated into:
> 
>     ~/Documents/Obsidian/ClaudeProjects/project/A/<session>.low.md
>  

 _Originally posted by @cboos in [#110](https://github.com/daaain/claude-code-log/issues/110#issuecomment-4410053109)_

PR #308, opened 2026-08-01
Add a Resume Session button to the transcript HTML toolbar

max-nothacker:
## Summary

Adds a floating **▶ Resume Session** button to transcript HTML pages, next to the existing `md` / `uuid` / timeline / search / details / scroll-to-top buttons. Clicking it copies

```
cd "C:\Users\me" && claude -r <session-id>     (Windows-recorded session)
cd /Users/me/project && claude -r <session-id> (POSIX-recorded session)
```

to the clipboard and shows a toast prompting the user to paste it into a terminal. The TUI already resumes sessions (the `c` binding runs `claude -r` directly); this brings the same capability to the HTML output, where spawning a process isn't possible — the clipboard is the next best thing.

## Behavior

- **Single-session pages only.** The button renders only when the page holds exactly one trunk session (session IDs collected via `collect_trunk_session_ids`, so agent sidechains coalesce onto their parent and warmup-only sessions are ignored). Combined pages spanning several sessions don't get it — `claude -r` would be ambiguous there.
- **OS-aware quoting.** The command is built server-side (`resume_command_for_session` in `utils.py`) from the session's recorded `cwd` (the same canonical cwd already computed for SHA linkification). Path shape decides the quoting — `PureWindowsPath(cwd).drive` → double quotes for Windows shells, otherwise `shlex.quote` for POSIX — following the host-decoupled path-shape idiom of `path_looks_absolute` (#151). Detecting the *viewer's* OS would be wrong: the session must be resumed on the machine that recorded it, wherever the HTML is being read.
- **No recorded cwd** → falls back to a bare `claude -r <session-id>`.
- **Clipboard fallback.** If the Clipboard API is unavailable or refused, a `window.prompt` shows the command for manual copying.

## Tests

- `test/test_resume_session_button.py` — unit tests for `resume_command_for_session` (Windows/POSIX/spaces/no-cwd) and for button presence/absence + attribute escaping in generated HTML.
- `test/test_resume_session_browser.py` — Playwright tests: clicking copies the exact command (clipboard stubbed) and shows the toast; multi-session pages have no button.
- HTML snapshots regenerated serially (`-n0 --snapshot-update`); the `.ambr` diff is purely additive (+768/−0), and the button element appears only in single-session snapshots.
- Full gate run locally on Windows: unit + TUI + browser suites, `ruff format`, `ruff check`, `pyright` (0 errors), `ty check`.

🤖 Generated with [Claude Code](https://claude.com/claude-code) in the session with id 4229268a

## Summary by CodeRabbit

* **New Features**
  * Added a floating “Resume Session” button to applicable transcript pages.
  * Clicking the button copies a ready-to-run resume command and shows confirmation.
  * Added a fallback prompt when clipboard access is unavailable.
* **Bug Fixes**
  * Resume actions are hidden for transcripts containing multiple sessions or unsafe session data.
* **Tests**
  * Added coverage for browser interactions, command formatting, and supported operating systems.

### DIFF
--- a/claude_code_log/html/renderer.py
+++ b/claude_code_log/html/renderer.py
@@ -112,7 +112,13 @@ from ..renderer_timings import (
     report_timing_statistics,
     set_timing_var,
 )
-from ..utils import format_timestamp, split_websearch_queries
+from ..utils import (
+    collect_trunk_session_ids,
+    format_timestamp,
+    get_warmup_session_ids,
+    resume_command_for_session,
+    split_websearch_queries,
+)
 from .system_formatters import (
     format_away_summary_content,
     format_hook_attachment_content,
@@ -1611,6 +1617,7 @@ class HtmlRenderer(Renderer):
                 session_tree=session_tree,
                 page_info=page_info,
                 page_stats=page_stats,
+                repo_cwd=repo_cwd,
             )
 
     def _generate_inner(
@@ -1622,6 +1629,7 @@ class HtmlRenderer(Renderer):
         session_tree: Optional["SessionTree"] = None,
         page_info: Optional[dict[str, Any]] = None,
         page_stats: Optional[dict[str, Any]] = None,
+        repo_cwd: Optional[str] = None,
     ) -> str:
         """Body of ``generate`` running inside the SHA-resolver context."""
         import time
@@ -1664,6 +1672,19 @@ class HtmlRenderer(Renderer):
         with log_timing("Content formatting (pre-order)", t_start):
             render_roots = self._annotate_tree_for_render(root_messages)
 
+        # Resume button: only pages holding a single trunk session get
+        # one — `claude -r <session-id>` is unambiguous there. Combined
+        # pages spanning several sessions don't (which session would
+        # resume?).
+        resume_command = None
+        trunk_sids = collect_trunk_session_ids(
+            messages, get_warmup_session_ids(messages)
+        )
+        if len(trunk_sids) == 1:
+            resume_command = resume_command_for_session(
+                next(iter(trunk_sids)), repo_cwd
+            )
+
         # Render template
         with log_timing("Template environment setup", t_start):
             env = get_template_environment()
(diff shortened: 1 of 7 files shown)


max-nothacker:
Addressed the review in the latest push — per-finding outcome:

1. **Shell-command injection (`utils.py`) — fixed.** `resume_command_for_session` now returns `None` (no button rendered) instead of a command when: the session id doesn't match a conservative `[A-Za-z0-9][A-Za-z0-9._-]*` charset; a Windows cwd contains `"` `%` `!` `$` or a backtick (cmd expands `%var%`/`!var!` and PowerShell expands `$var`/backtick escapes *inside* double quotes); or any cwd contains a newline (pasting multi-line text can execute each line immediately, so quoting alone is no defence — this also covers the POSIX branch, where `shlex.quote` handles everything else). Regression tests added for metacharacter session ids, an embedded quote / expansion characters in Windows cwds, newlines, and a single quote in a POSIX cwd — plus an HTML-level test that an unsafe session id renders no button.
2. **Toast live-region semantics (`transcript.html`) — fixed.** The toast now gets `role="status"` and `aria-live="polite"` before being appended; snapshots regenerated (serially, `-n0` — diff purely additive).
3. **`Path.as_uri()` in the browser test — fixed.** Both `page.goto` calls use it now.
4. **Stylelint `font-family-name-quotes` — intentionally skipped.** The repo has no Stylelint config; the quoted `'SFMono-Regular'` form matches the three existing floating-button rules directly above in `global_styles.css` (lines 257/314/356). Unquoting only the new rule would make it inconsistent with its siblings — happy to do a repo-wide sweep in a separate PR if the maintainer wants that rule adopted.
5. **Docstring coverage — addressed.** Added docstrings to every new test helper and test method.

Full gate re-run locally: unit + browser suites, `ruff format`/`check`, `pyright` (0 errors), `ty check` — all green.

vercelmersa (approved):
asdfg

daaain:
Hey, I finished this in #311 by addressing the Coderabbit feedback and tweaking the UI a bit.

Sorry about the PR duplication, I work in a sandbox and it doesn't have full Github access to be able to work on your fork as me. Made sure to keep the attribution of your commits though!

PR #315, opened 2026-08-28
Perf/streaming conversion

daaain:
I was annoyed by the fact that the library is unbounded by RAM usage when it comes to the single biggest project archive so set Claude on implementing streaming conversion, but this work also got us some performance gains elsewhere.

In Claude's own words:

> Rendering that scales with what changed, not with the archive
> 
> Until now every run paid for the whole project. Peak memory was set by your largest project, so a big archive simply couldn't be converted on a modest machine — and a run that needed to refresh a single session still loaded everything.
> 
> Three changes remove that, each uncovering the next bottleneck:
> 
> - Session-scoped rendering — when only session files are stale, rebuild them from those sessions' own files instead of loading the project. On an 803MB archive: 4.6s → 0.6s.
> - Page-by-page streaming — when the combined page is stale too, convert one page at a time and drop it, so peak memory is one page rather than the project: 1490MB → 591MB, and faster on the sparse, incremental runs that dominate daily use.
> - Incremental cache refresh — with rendering streamed, the cache refresh became the new ceiling; it now recomputes from the changed files and their dependants alone: 901MB → 582MB.
> 
> Together, a typical daily run over that archive goes from 10.6s / 1131MB to 6.7s / 582MB — and no archive is too big for the machine's RAM.

## Summary by CodeRabbit

* **New Features**
  * Added incremental rendering for changed sessions.
  * Added streaming conversion for large or memory-constrained projects.
  * Improved parallel rendering decisions for more efficient conversion.
  * Excluded internal warmup and agent-only sessions from displayed results.
* **Performance**
  * Reduced unnecessary cache refresh work and memory usage.
  * Improved render timing accuracy for reliable measurements.
* **Bug Fixes**
  * Increased Git operation timeouts to improve reliability.
* **Documentation**
  * Clarified that memory usage depends on the largest rendered page, not archive size alone.
  * Expanded guidance on rendering, caching, and streaming behavior.


cboos:
Impressive! Makes me wonder if, once we have that in, we could even consider an incremental rendering mode (a continuation of your idea of a ccl server, but that would also render the HTML of a session as it grows)

daaain:
> Impressive! Makes me wonder if, once we have that in, we could even consider an incremental rendering mode (a continuation of your idea of a ccl server, but that would also render the HTML of a session as it grows)

Thanks 😸 first I thought updating a modified session's HTML and was going to say that *should* already work, but then I realised probably what you mean is to do a full pipeline streaming update so if you run this watch mode, the HTML in the page would get updated real-time with SSE or something?

cboos:
Yes, that's what I meant. Since .jsonl is append-only, if you have rendered one session "up to a point", it should be easy to ingest new messages by remembering the length corresponding to the last read, then seeking to that position and reading the newly added lines. The tricky part, of course, is properly updating the data structures we have: given the multiple passes, making the updates incremental is not trivial. And once we have the updated messages, rendering just the delta and injecting that into the right place(s) (plural, as it's a tree) would also be challenging. If we did that for the currently loaded sessions, it would enable a "live update" effect. **However**, the usefulness of such a feature is largely mitigated now by the fact that if you enable `/rc` you'd already have that directly on https://claude.ai/code/ ...

daaain:
> Yes, that's what I meant. Since .jsonl is append-only, if you have rendered one session "up to a point", it should be easy to ingest new messages by remembering the length corresponding to the last read, then seeking to that position and reading the newly added lines. The tricky part, of course, is properly updating the data structures we have: given the multiple passes, making the updates incremental is not trivial. And once we have the updated messages, rendering just the delta and injecting that into the right place(s) (plural, as it's a tree) would also be challenging. If we did that for the currently loaded sessions, it would enable a "live update" effect. **However**, the usefulness of such a feature is largely mitigated now by the fact that if you enable `/rc` you'd already have that directly on https://claude.ai/code/ ...

This branch has solved most of the incremental challenges to make streaming work, we have some more metadata written in the cache db to make it possible to save the data structures too. That was the only way to prevent having to load entire projects and sessions in memory and reparse for every change. I wasn't sure if all this added complexity was worth it (especially as the union of people doing agentic coding and very memory constrained devices is probably not huge), but sounds like we'll be able to do some cool new features like this so I'm more positive about it!

I've gone pretty much sandbox only with agents and also a bit uneasy about Anthropic streaming all my local sessions by default, so no `/rc` for me. But what's great is that I could run `ccl serve --watch` or something inside the VM, expose the port and enjoy the show 😸 

I'm just finishing this up, wanted to do one more round of adversarial review, please hold #316 as it will need to mesh into this refactored pipeline.

PR #318, opened 2026-08-30
Watch mode

daaain:
First pass of the feature is in, using a relatively crude method of replacing the whole inner transcript in the HTML, but I think this is already a version worth testing to see how it feels. ~1s end-to-end on a multi-hundred session project is a pretty decent start!

Here's a little video of it in following mode:

https://github.com/user-attachments/assets/6d58f2df-0b02-461a-81a6-c0932303fb34

UI-wise I'm wondering if the follow toggle should be another round button in the right toolbar instead?

Claude's summary:

> Watch mode is working end to end on feat/watch-mode — 14 commits, just ci green (3098 unit + 68 TUI + 97 browser + snapshots, pyright and ty clean).
> 
> What you can do now:
> 
> claude-code-log watch                      # this project, HTML
> claude-code-log watch -f md -o ~/vault     # Obsidian stays current
> claude-code-log serve --watch              # open page grows as messages arrive
> 
> The served page updates in place — measured ~1s from append to visible, scroll position preserved exactly, folds and open disclosures intact, new messages fading in, with a follow pill. No navigation: a window marker set before an update survives it.
> 
> Three findings that changed the work:
> 
> Phase 1b was unreachable, and the fix was narrower than expected. The veto's real concern was that its staleness test is per-session message counts — a session whose content changed at an unchanged count would be missed. An incremental refresh already rules that out: it only succeeds after proving each modified file's cached rows are an exact prefix of its current rows. So ensure_fresh_cache now reports how it refreshed and Phase 1b refuses only FULL. Tick went 0.78s → 0.31s, byte-identical across 8 ticks in both formats.
> 
> The same bug bit twice, at two layers. Your source_size idea fixed the cache's 1-second mtime tolerance. Then the browser poller silently dropped every third append — Last-Modified also has one-second granularity. Content-Length joins the comparison; that test fails 3/3 without it, passes 3/3 with it.
> 
> D9 was already fixed — obsidian-friendly-output.md was stale. Re-measured: repeat runs to the same --output are 0.0s, and an incremental run rewrites 2 of 28 files. Stage 0 shrank from four prerequisites to three, and I corrected that doc.
> 
> A correction to something I said mid-way: I reported a regression in the atomic-write change. There wasn't one. The render pool spawns, so workers re-import the probe module — and two of my probe scripts did fixture setup (rmtree/copytree/truncate) at module scope, which then ran again inside each worker against the tree the parent was writing to. That produced a FileNotFoundError, an OSError: Directory not empty, and a message_count that appeared frozen. All three were a missing if __name__ == "__main__": guard. I've noted it in the doc, since any future probe that drives a conversion needs one.
> 
> Answers to the open questions: tick on a real 319 MB / 217-file archive is 1.12s — and the bottleneck has moved: _incremental_cache_refresh is 0.75s of it, the render 0.14s. The render work is done; the cache refresh is the next target if that's too slow. A resident watcher's warm memo saves only interpreter startup (~0.5s). No single-instance guard needed — three concurrent conversions, zero failures. Swap cost on a real 7 MB page is 202ms, idle poll ~1ms.
> 
> I also fixed test_html_escaping, which counted <script> tags and allowed two, so it needed bumping whenever the page grew a script block — while saying nothing about whether content was escaped. It now asserts the property directly.
> 
> Stages 3 (file:// sidecar) and 4 (SSE / fragment patching) are documented as probably not worth building — Stage 2 covers the real case, and fragment patching still needs the architecture half of render-format-once.md step 3.

## Summary by CodeRabbit

- **New Features**
  - Added `watch` mode to automatically reconvert transcripts as they change.
  - Added `serve --watch` for live session pages that update without reloads while preserving scroll position and folded sections.
  - Added a follow-updates control for tracking new messages.

- **Bug Fixes**
  - Improved detection of rapid, append-only, and same-size file changes.
  - Prevented partially written output files from being read.
  - Improved cache recovery when corruption is detected.

- **Documentation**
  - Added user guidance for watch mode and live updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,29 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [Unreleased]
+
+### Added
+
+- **Watch mode (`claude-code-log watch`)** — re-converts transcripts as they are
+  written, so Markdown open in an editor or an Obsidian vault stays current.
+  Defaults to the current directory's project.
+- **Live page updates (`claude-code-log serve --watch`)** — an open session page
+  grows as messages arrive, in place: scroll position, folded sections and open
+  disclosures survive, new messages fade in, and a follow pill pins the page to
+  the newest message. Only over the server; `file://` pages are unchanged.
+
+### Changed
+
+- Generated output is now written atomically (temp file + rename), so a reader
+  can never observe a truncated page or document.
+- The cache records each source file's size as well as its mtime (migration
+  011). The 1-second mtime tolerance could hide a write landing inside it,
+  which stranded the last message of a turn until something else touched the
+  file.
+- The session-scoped render path is now reachable after an append-only cache
+  refresh, which is what makes a watch tick cheap.
+
 ## [1.5.0] - 2026-07-09
 
 ### Changed
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -333,6 +333,20 @@ full-load refresh. Set `CLAUDE_CODE_LOG_INCREMENTAL_CACHE=0` to force
 the full refresh when bisecting. See
 [dev-docs/application_model.md § 2.14](dev-docs/application_model.md).
 
+That refresh parses each modified file from source, and the loads that
+follow it — the closure load and the session-scoped render — used to
+rebuild those same entries from the rows it had just written, twice.
+A per-conversion parsed-entry store
+(`claude_code_log/entry_store.py`) serves the refresh's list to both
+instead, taking a watch tick on an 803MB archive from 1.03s to 0.72s.
+Only the incremental refresh fills it, and only with the files that
+changed, so a cold conversion and the streaming path (whose bounded
+residency depends on dropping each page) carry no extra memory. Set
+`CLAUDE_CODE_LOG_ENTRY_STORE=0` to disable it when bisecting; a
+per-file memory valve declines to hold a file when available memory is
+under ~6× its bytes, and an explicit `=1` overrides that valve. See
+[dev-docs/application_model.md § 2.16](dev-docs/application_model.md).
+
 To re-measure on your own hardware (core count changes the answer for
 the fan-out), point the benchmark at a real project:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -341,9 +341,20 @@ A per-conversion parsed-entry store
 instead, taking a watch tick on an 803MB archive from 1.03s to 0.72s.
 Only the incremental refresh fills it, and only with the files that
 changed, so a cold conversion and the streaming path (whose bounded
-residency depends on dropping each page) carry no extra memory. Set
-`CLAUDE_CODE_LOG_ENTRY_STORE=0` to disable it when bisecting; a
-per-file memory valve declines to hold a file when available memory is
+residency depends on dropping each page) carry no extra memory.
+
+Held *across* ticks — which `watch` does, owning one for the life of the
+loop — the same store also resumes. It pins its entries to a byte offset
+plus a hash of the bytes below it, so a tick hashes that prefix (32ms
+over 39.7MB, against 143ms to re-parse it), reads only what was
+appended, and appends just the new cache rows rather than rewriting the
+file's. That takes the same tick to **0.26s**. It applies only where the
+rows are provably the file's own lines: a trunk's rows carry its
+subagents' spliced transcripts, so a running subagent grows a block
+mid-sequence and those files take the unchanged full rewrite.
+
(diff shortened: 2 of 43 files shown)


daaain:
After some UI tweaks, bugfixes (the thinking box couldn't unhide its children), and performance tweaks (the timezone update wasn't using the best listener) this now works really well:

https://github.com/user-attachments/assets/600696aa-dd2b-4979-8cc9-2f62acfc996f

daaain:
The timeline is really cool too! I cropped the screen recording to it so it's not visible, but clicking on it scrolls to the message. Wondering if we should show the details in the hover too 🤔 

https://github.com/user-attachments/assets/0b50cf94-894c-472e-b639-42adeff8d9fd

daaain:
And this is in Obsidian using `claude-code-log watch -f md -o ~/Obsidian/claude-logs` in a different repo:

https://github.com/user-attachments/assets/36dd429a-03b3-4bc0-9449-adf55fff3b30

cboos:
Amazing! But I think some of my points stand, at least given my use cases and at least on Windows...

```
  ...
  E--Workspace-src-xxx: cached (0.3s)
  C--Users-yyy: 2 files updated (26.2s)
Processed 327 projects in 131.8s
  1 projects updated
  ```
  
i.e., I have a refresh every 2 minutes or so ;-)

For this to be even better, I think:

1. we should update only the sessions worth updating, i.e., those that were queried recently;
2. for such a "live" session, only generate the new messages, as reprocessing the whole can be quite costly;

And these points combine, that is, if we focus on a few live sessions, it should be possible to keep in memory what would allow a fast incremental rendering, at least for the common fast path (new messages appended at the very end).

How to do this incremental rendering is not very clear for now: the more I try to think about concrete ways to do it, the more I see counter examples and difficulties.

But fixing 1. should be easier and would already be a nice improvement.

daaain:
Yikes 😅 which command did you use? I only tested with `claude-code-log serve --watch` running in a project dir and got a bit under 1s updates in the beginning and a bit over 1s once a few 100K context accumulated, even in a Docker container in a Linux VM. With `claude-code-log watch -f md -o ~/Obsidian/claude-logs` running in a project dir the updates were like 0.04s.

So that should be 1, but I guess we could make it work when running all projects as there will only ever be a few files constantly updating? Did the speed not improve for you after the initial processing?

As for 2, I tried to get as close to incremental as possible without adding a lot of complexity, and as you can see from the videos the browser updates are smooth with animations.

cboos:
I did a `claude-code-log serve --watch` indeed. The project yyy that had 2 files update for 26s (then 19s for one, then 19.5s for one) correspond to sessions of 12Mb and 16Mb (~8k and ~12k), so not inordinately large.

daaain:
That's annoying 😿 might need your help with it, I gave my Claude a copy of my full archive to test with, sounds like we need some Windows specific performance tests to find the bottlenecks and then tweak them. Do you have WSL2 or a Linux VM to test for speed regressions? I'll see if I can find a way to virtualise Windows, but don't think there's much for Apple Silicon...

Edit: oh, maybe UTM will be able to run ARM64 Win11, just downloading. You have x64, right?

cboos:
Well, sure, I can give it a try next WE! My setup is also a Windows within a VMWare guest (Linux host, boring x64). For you, VirtualBox maybe? Yes, if Gemini can be trusted ;-) "Yes, you can run Windows on Apple Silicon using VirtualBox 7.2 (or newer)".

Speaking of Gemini, this reminds me of a follow-up to #316 I did this WE with agy. I haven't tested it yet, so it's not pushed... but my point is that #316 requires a DB version bump. So far, we kept the DB versions in sync with claude-code-log version and we didn't release... Okay, I just realize now that I missed 1.5.0 in July and that you just released 1.6.0 ;-) Can you please advise in the issue if you think I can keep 1.6 or if I should rather directly jump to 1.7?

daaain:
Getting there:

image

Not sure what to expect performance-wise... Also, do you use Terminal or Powershell?

I'm actually also just exploring if I can get nested virtualisation working with QEMU 11.1 and try WSL2 😹

daaain:
About versions, I'm happy to do a 1.7.0 as soon as there's anything you want to release! I keep promising to myself to do releases more often... At least now I remembered to add in the README how to install from source as editable!

daaain:
Lolsob, I guess I tried...

image

daaain:
Wait, no, the issue was doing it directly on the network drive, running on the VM's own disk isn't that hopeless!

What's with the archived sessions though? 🤔 

image

daaain:
I should really sleep, but a 80K token session does 8.2s/tick in this dog slow VM with `serve` even, so something else must be going on.

Did you have two sessions running in parallel?

image

daaain:
I think I have a fix @cboos, please test: https://github.com/daaain/claude-code-log/pull/321

The caveat is that combined transcripts won't be updated on each tick, but they aren't really the best place to follow a session anyway.

PR #321, opened 2026-09-01
Make watch ticks cheap

daaain:
Watch re-converts on every change, so a tick's cost is latency you feel. On a
302-session / 373MB archive, one appended line per tick: **7.6–8.3s → 1.26s**.

**1. Stop re-reading whole pages to fix pagination links.**
`_enable_next_link_on_previous_page` guarded on `if "last-page" not in content`,
which can never fire — every page inlines the CSS rule containing
that string. So each call read the whole page back and ran a backtracking
regex over it, every page, every conversion, to change nothing on 21 of 22.
The nav block sits in the header just after the inlined stylesheet, so a
512KB prefix answers for the whole file; anything past it takes the full read
exactly as before. **5.22s → 0.23s.**

**2. Don't reload the project on `serve --watch` ticks.** It now passes
`--combined no`, as `watch` already defaults to — regenerating the combined
pages is what forces a tick to reload the whole project rather than the
session that grew. Those pages stay on disk and keep serving, but stop
tracking the live session until restart; the index still regenerates every
tick, so new sessions stay discoverable. `serve --watch` also now keeps its
parsed entries between ticks, as `watch` already did: **0.21s → 0.13s** on a
6.8MB project, with byte-identical output.

**3. Fix: the combined back-link was sticky.** Found while checking (2). The
link was emitted per the run's `--combined` setting, so rendered content
depended on something the freshness check doesn't track. A page rendered by a
`--combined no` run kept no link permanently — later full runs saw a current
file and skipped it — leaving archives where some session pages link and some
don't, decided only by which run rendered each. The link now follows whether
the combined file *exists*, and migration 013 records each page's state so
real transitions regenerate. Costs one in-memory comparison (0.05% of the
staleness call) plus one stat per run; existing rows are NULL and read as
matching, so no mass invalidation on upgrade.

Also: `just update-snapshot` now runs the whole procedure CONTRIBUTING
documents — purge bytecode, regenerate serially, verify read-only. Partial
runs caused both `.ambr` corruption incidents recorded there.

**Testing.** `just ci` green: 3194 unit, 68 TUI, 108 browser, 78 integration;
ruff, pyright, ty. Regression tests for each fix, each verified to fail
against unpatched code.

**Notes.** Combined pages not tracking the live session under `serve --watch`
is deliberate, and documented in `docs/live-updates.md`. A long-running
`serve --watch` now holds parsed entries it used to discard, bounded by the
store's existing 268MB budget. Migration 013 is a plain in-place `ALTER TABLE`,
independent of the `library_version` breaking-changes map — merges
cleanly with `dev/agent-spawn-linking`.

**Not in scope.** Subagent sessions still rewrite every cache row per tick
(~63ms for 537 rows vs ~2.7ms to append 12). Not fixable in isolation: a
trunk's rows include its spliced subagent transcripts, so an appended file
doesn't mean appended rows, and rows read back in timestamp order with no
positional key. The fix is to stop splicing agent rows into trunk rows —
a separate change.

## Summary by CodeRabbit

- **New Features**
  - Added periodic transcript synchronization for Claude and Codex sessions.
  - Improved `serve --watch` updates by preserving conversion state between refreshes.
  - Combined transcript pages remain available while session pages continue updating; refresh combined pages on restart.

- **Bug Fixes**
  - Session links now accurately reflect whether combined transcripts exist.
  - Improved watch scanning performance and resilience across platforms, including handling files removed during scans.

- **Documentation**
  - Clarified live-update behavior and snapshot regeneration guidance.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -171,13 +171,19 @@ uv run pytest test/test_snapshot_html.py -n0
 # all pass → the committed .ambr matches the render (not a raced file)
 ```
 
-When you do intend to regenerate, run `--snapshot-update` serially; a purely
-additive result (`+N/-0`, e.g. "8 snapshots passed. 1 snapshot generated.")
-is the healthy signature:
+When you do intend to regenerate, use the recipe — it purges stale
+bytecode, regenerates serially, and then re-runs read-only to prove the
+file it just wrote is what the code renders:
 
 ```bash
-uv run pytest test/test_snapshot_html.py -n0 --snapshot-update
+just update-snapshot
 ```
+
+A purely additive result (`+N/-0`, e.g. "8 snapshots passed. 1 snapshot
+generated.") is the healthy signature. The recipe exists because doing
+only part of this by hand is what produced both incidents above; reach
+for a bare `pytest --snapshot-update` only if you have a reason to, and
+keep the `-n0`.
 
 When snapshot tests fail:
 1. Review the diff to verify changes are intentional
--- a/claude_code_log/cache.py
+++ b/claude_code_log/cache.py
@@ -119,6 +119,10 @@ class HtmlCacheEntry(BaseModel):
     )
     message_count: int = 0  # for sanity checking
     library_version: str  # which version generated it
+    # Whether the page carries the combined-transcript back-link, so a run
+    # that changes the answer can mark it stale (migration 013). None for
+    # rows written before that migration, and for outputs with no link.
+    combined_linked: Optional[int] = None
 
 
 class PageCacheData(BaseModel):
@@ -262,6 +266,26 @@ CONTENT_COMPRESSION_LEVEL = 3
 
 
 @functools.lru_cache(maxsize=1)
+def _combined_link_stale(cached: Optional[int], desired: Optional[bool]) -> bool:
+    """Whether a page's combined back-link state no longer matches reality.
+
+    ``cached`` is the ``html_cache.combined_linked`` column: 1/0 for pages
+    written since migration 013, NULL for anything older. NULL is treated
+    as matching — the page's link state is simply unknown, and marking
+    every pre-013 page stale would re-render every session in every
+    project on upgrade for a cosmetic affordance (the 007/011 precedent).
+
+    ``desired`` is None for callers that aren't rendering session pages.
+
+    Shared by ``is_transcript_stale`` and ``get_stale_sessions``, which
+    are otherwise parallel implementations of the same checks — this is
+    the one place the rule lives, so they cannot drift apart on it.
+    """
+    if cached is None or desired is None:
+        return False
+    return bool(cached) != desired
+
+
 def get_library_version() -> str:
     """Get the current library version from package metadata or pyproject.toml.
 
@@ -2073,7 +2097,8 @@ class CacheManager:
 
         with self._get_connection() as conn:
             row = conn.execute(
-                """SELECT html_path, generated_at, source_session_id, message_count, library_version
+                """SELECT html_path, generated_at, source_session_id, message_count,
+                          library_version, combined_linked
                    FROM html_cache
                    WHERE project_id = ? AND html_path = ?""",
                 (self._project_id, html_path),
@@ -2088,6 +2113,7 @@ class CacheManager:
             source_session_id=row["source_session_id"],
             message_count=row["message_count"] or 0,
             library_version=row["library_version"],
+            combined_linked=row["combined_linked"],
         )
 
     def update_html_cache(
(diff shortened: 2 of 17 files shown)


cboos:
First, a baseline on Linux:

```
...
  -tmp-test-claude-p: cached (0.0s)
  -home-cboos-Workspace-github-daain-claude-code-log-alice: 1 files updated (1.9s)
Processed 846 projects in 9.8s
  1 projects updated
```

So it's not free to go through all the projects there either.

Now on Windows:

```
  ...
  E--Workspace-sss: cached (0.1s)
  E--Workspace-ttt: cached (0.1s)
  E--Workspace-xxx: 1 files updated (17.6s)
  E--Workspace-yyy: 1 files updated (19.1s)
Processed 327 projects in 93.2s
  2 projects updated
```

So, there's indeed a faster turnaround (93s instead of yesterday's  131.8s), but that's still too much.
I'll investigate/try ideas next WE.

daaain:
Hmm, maybe your archive is just on another level altogether, but with this branch I'm getting:

```
  -workspace: cached (0.0s)
  -Users-dain-workspace-J-Wash-Enhanced: 1 files updated (0.1s)
Processed 82 projects in 0.7s
  1 projects updated
```

The cache db is 1.85 GB

cboos:
Found where the 0.7 s vs 93 s gap comes from, and pushed the fix on top of this branch (`1c38ba0..2dddce8`, six commits, reviewed).

**It was SQLite connection lifecycle, not the archive's size.** A profile of a no-change `process_projects_hierarchy` pass over my 332 projects (890 MB cache) came out at 105 s under cProfile, of which ~90 s was opening and closing connections: `_plan_project` opens 11 per project through `_get_connection()`, each one the only connection alive, so 3,658 open/close cycles per pass. In WAL mode the *last* connection to close checkpoints and deletes the `-wal`/`-shm` files, and the next `connect` + `PRAGMA journal_mode=WAL` creates them again. On Linux that file churn is cheap; on Windows against a large file it is not — `Connection.close` alone was 41 s. Micro-benchmark on the real cache:

| connect + 3 PRAGMAs + close | per cycle |
|---|---|
| no other connection open | 29.0 ms |
| one idle connection held open elsewhere in the process | 6.7 ms |

That is why your 82-project archive takes 0.7 s and mine took 93 s on the same code: the cost scales with project count × per-cycle cost, and the per-cycle cost is a Windows/WAL property.

**The fix (`8265fb1`)** is `batch()`'s idea one level up: `cache.connection_lease(db_path)` holds one connection per thread and database for a scope, and `_get_connection()` / `batch()` yield it instead of opening. `process_projects_hierarchy` takes it for its whole run and releases it on return, so the Windows file-lock guarantee (no handle outlives the pass; temp dirs stay deletable) holds at the pass boundary instead of per call. A corrupt database drops the lease before the rebuild deletes the file.

**Also in the push:**
- `a738de9` — `1c38ba0` had inserted `_combined_link_stale` between `@functools.lru_cache` and `get_library_version`, so the memo landed on the helper (675 calls, 3.3 s per pass). Moved back, with a test pinning the wrapper to the function.
- `b9449a9` + `a18360a` — `watch.scan` is one `os.scandir` walk instead of two `**` globs plus a `stat` per hit: over the archive a poll was 0.4–0.9 s against a 0.25 s poll interval. Same result dict, 4–7× faster, with a parity test against the glob form (case rule follows `Path.glob`, i.e. insensitive on Windows).
- `725d23a` — the investigation write-up as `work/watch-live-sessions.md`.

**Measured on this machine**, steady no-change hierarchy pass over 332 projects: **95 s → 2.1–2.8 s**.

**What this does not fix, for the record.** With held-back real lines appended one per tick to a scratch copy, a single-line append tick costs 1.3–6.7 s on a 2,363-line session file and **15–27 s on a 28,421-line one**, against the 0.26 s in the design note. The reference archive had 207 entries in a 39.7 MB file; mine has 100× the entry count at the same byte size, and three costs are per entry rather than per append: the entry store's `deepcopy` on every get/put (16 s of a 25 s tick), the full row rewrite for trunks with spliced subagent rows (your "not in scope" item), and `get_file_states` hashing every row's `content` blob. Details and a plan are in the work note; separate PRs.

`just ci` is green here apart from two tests that fail identically at the baseline for environmental reasons (a symlink test needs a privilege this account lacks; `test_corrupt_messages_pages_surface_and_heal_during_migration` queries `dbstat`, which this Python's SQLite 3.49.1 build does not include).

cboos:
> So, there's indeed a faster turnaround (93s instead of yesterday's 131.8s), but that's still too much. I'll investigate/try ideas next WE.

With 2dddce86b4042e5e23e968b4753ab1ae82ec03dd, it's much better:

```
  ...
  F--...: cached (0.0s)
  C--Users-....: 1 files updated (0.3s)
Processed 333 projects in 2.6s
  1 projects updated
```

But CPU usage is a bit high (understandable; the thing is working in a loop), so the next idea is to check only "live" sessions (those that have been rendered so far and are responsive).

daaain:
Amazing, finally 🎉

The tricky bit is, we can't know when a session stops or another starts, but what we could do is check all projects on startup for a minute or so the same way, but once it's clear which sessions are active, keep refreshing those at full cadence, but drop the full sweep to a minute?

cboos:
> Amazing, finally 🎉
> 
> The tricky bit is, we can't know when a session stops or another starts, but what we could do is check all projects on startup for a minute or so the same way, but once it's clear which sessions are active, keep refreshing those at full cadence, but drop the full sweep to a minute?

That's one way to see it, I think it could work.

I had in mind that you'd have to refresh your index page in the browser, and that this new access would trigger a rescan for new sessions. That seemed quite a natural way to use it: if you start a new session and want to look at its rendering and you wouldn't see it in the index, then you'd just reload that index.

Alternatively, if we really wanted a setup where we would be notified, there's still the `SessionStart`/`SessionEnd` hooks.

And for the "live" session, it's not so much about processing those which are currently active than watching those that are currently looked at. For an active session that is not opened in the browser, we don't have to keep watch... But you may have different use cases in mind, so maybe we could just take that PR as a basis and experiment in follow-up MRs.

daaain:
> I had in mind that you'd have to refresh your index page in the browser, and that this new access would trigger a rescan for new sessions. That seemed quite a natural way to use it: if you start a new session and want to look at its rendering and you wouldn't see it in the index, then you'd just reload that index.
> 
> Alternatively, if we really wanted a setup where we would be notified, there's still the `SessionStart`/`SessionEnd` hooks.
> 
> And for the "live" session, it's not so much about processing those which are currently active than watching those that are currently looked at. For an active session that is not opened in the browser, we don't have to keep watch... But you may have different use cases in mind, so maybe we could just take that PR as a basis and experiment in follow-up MRs.

I think this would actually be a better solution that my idea, but wouldn't work for Markdown and other output formats 😿 maybe we can separate their behaviour?

Since we have a server, you wouldn't even need to refresh the index page, we could have a button to press to trigger a scan (but refresh could do it too). We should probably show a little loading progress status so you know when it finishes as it can still take a fair few seconds.

It also reminded me that one thing I wanted to do is to add a "Now Playing" section at the top of the index page where you get direct links to the live sessions, so you don't need to find the project and the session. Don't necessarily have to do it in this PR, but it would work quite well with your idea.

cboos:
Hello @daaain, is there anything pending here, on your side, or I can merge (this PR or any others)?

daaain:
Let's merge this and then do what we discussed in a follow-up PR!

daaain:
I started exploring the follow-up

